PR #13, opened 2022-07-19
Initial dashboard controller

HubertStefanski:


addreas:
I forgot to mention it before, but since you mentioned deletion being a next step you might be interested in the finalizer bits here for inspiration: https://github.com/grafana-operator/grafana-operator-experimental/pull/14/files#diff-b27f952cb9cb056c639f702464d779a287cd0e37178eb3cfff1b6d45b058b554R90

NissesSenap:
So i haven't looked at the code or the need. But in general if we can do everything without finalizes i think that should be the goal. 
They are very painful to manage from an operation point of view. 
Hurts allot when you try to delete the operator and in general a pain to take in to account. 

Could be something to talk about during next meeting

Makefile
```diff
@@ -118,13 +118,35 @@ undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/confi
...
-controller-gen: ## Download controller-gen locally if necessary.
-	$(call go-get-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen@v0.7.0)
+controller-gen:
+ifeq (, $(shell which controller-gen))
+		@{ \
+		set -e ;\
+		CONTROLLER_GEN_TMP_DIR=$$(mktemp -d) ;\
+		cd $$CONTROLLER_GEN_TMP_DIR ;\
+		go mod init tmp ;\
+		go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.9.2 ;\
```
pb82, line 128: what version of the operator-sdk are you using? This bumps controller-gen to a newer version which seems to conflict with some of the other arguments. When running `make manifests` I get:

```
Error: unable to parse option "crd:maxDescLen=0,generateEmbeddedObjectMeta=false": [unknown argument "generateEmbeddedObjectMeta" (at <input>:1:40) extra arguments provided: "false" (at <input>:1:41)]
```

HubertStefanski: Ahh, my bad, I didn't realize I was running v1.22, which is where I fixed the issue I was running into with kubebuilder previously, I'll downgrade this today and see if I can revert this change to work with v1.13

controllers/dashboard_controller.go
```diff
@@ -0,0 +1,329 @@
...
+				//URL:       "",
+			},
+			Model: dashboardFromJson,
+			//Folder:    0,
+			Overwrite: false,
+			Message:   "",
+		})
+
+		if err != nil {
+			if !strings.Contains(fmt.Sprintf("%s", err), statusCodeDashboardExistsInFolder) {
```
weisdd, line 223: Since `error` is an interface with `Error()` method that returns its text representation ([source code](https://cs.opensource.google/go/go/+/refs/tags/go1.18.4:src/builtin/builtin.go;drc=f524a2baa4aca9e603fbf3891951b53b497dc0a1;l=268)), you might prefer something like:
```go
if !strings.Contains(err.Error(), statusCodeDashboardExistsInFolder)
```
instead of:
```go
if !strings.Contains(fmt.Sprintf("%s", err), statusCodeDashboardExistsInFolder)
```

Also, since there will be lots and lots of places where we'll need to inspect particular error codes (e.g. 2 places just in your PR), I'd suggest to turn it all into a little helper. Plus, to make sure that we always inspect status code, not body contents of the error (maybe I'm too careful with this, and numbers never appear there), I'll make the helper close to grafana api client error format, which is:
```go
if resp.StatusCode >= 400 {
    return fmt.Errorf("status: %d, body: %v", resp.StatusCode, string(bodyContents))
}
```
([source](https://github.com/grafana/grafana-api-golang-client/blob/9c38b6f437bb9688f655ace9e66cc8b8f91ce0f4/client.go#L119))

An example of how it could be implemented (if you like the idea ofc):
```go
func errorStatusCodeIs(err error, code string) bool {
	prefix := fmt.Sprintf("status: %s,", code)
	return strings.HasPrefix(err.Error(), prefix)
}
```

controllers/client/grafana_client.go
```diff
@@ -88,18 +57,26 @@ func NewGrafanaClient(ctx context.Context, c client.Client, grafana *v1beta1.Gra
...
+		Client: &http.Client{
 			Transport: transport,
 			Timeout:   time.Second * timeoutSeconds,
 		},
-	}, nil
-}
+		// TODO populate me
+		OrgID: 0,
+		// TODO populate me
+		NumRetries: 0,
```
weisdd, line 73: Just a comment:
Their library has a hardcoded value for delays between retries:
```go
// Wait a bit if that's not the first request
if n != 0 {
	time.Sleep(time.Second * 5)
}
```
([source](https://github.com/grafana/grafana-api-golang-client/blob/9c38b6f437bb9688f655ace9e66cc8b8f91ce0f4/client.go#L81))
5 sec is fine for terraform, because everything is taking ages there, but it's is too long for operators as they are expected to do many things in a timely manner. So, we'll need to either write our own wrapper that will implement retry logic or to try to prepare a PR to the upstream library that would make retry timeout more flexible.

controllers/dashboard_controller.go
```diff
@@ -0,0 +1,329 @@
...
+				if resp != nil {
+					status.UIDforInstance[grafanaNamespacedName] = resp.UID
+				}
+				return err
+			}
+		}
+		if resp != nil {
```
pb82, line 230: should we also check the response status here?

controllers/dashboard_controller.go
```diff
@@ -0,0 +1,329 @@
...
+			for grafanaInstance, dashboardUID := range status.UIDforInstance {
+				if grafanaInstance == namespacedName {
+					return dashboardUID
+				}
+			}
+		}
+		return ""
+	}
+
+	grafanaNamespacedName := fmt.Sprintf("%s%s", grafana.Namespace, grafana.Name)
```
pb82, line 176: we can add a `/` between the namespace and the name for better readability

controllers/dashboard_controller.go
```diff
@@ -0,0 +1,329 @@
...
+			r.Log.Error(err, "could not delete dashboard from instance", "instance", grafanaNamespacedName)
+			return err
+		} else {
+			delete(status.UIDforInstance, grafanaNamespacedName)
+
+			if err := r.reconcileDashboardStatus(cr, status); err != nil {
+				r.Log.Error(err, "error while updating dashboard status")
+				return err
+			}
+			if status == nil {
```
pb82, line 200: when would the status be nil? When the last dashboard is removed, the map is empty, but not nil. Is a `len(status) == 0` missing here?

controllers/dashboard_controller.go
pb82: +1 on checking the status code instead of the string representation

controllers/dashboard_controller.go
```diff
@@ -0,0 +1,329 @@
...
+	}
+
+	grafanaNamespacedName := fmt.Sprintf("%s%s", grafana.Namespace, grafana.Name)
+	dashUIDfromStatus := getDashUIDFromStatus(grafanaNamespacedName)
+
+	if cr.DeletionTimestamp != nil {
+		r.Log.Info("found a deletion timestamp for dashboard", "dashboard", cr.ObjectMeta.Name, "instance", grafanaNamespacedName)
+		if err := grafanaClient.DeleteDashboardByUID(dashUIDfromStatus); strings.Contains(fmt.Sprintf("%v", err), "404") {
+			r.Log.Info("the dashboard couldn't be found on the instance, nothing to do, removing finalizer")
+			controllerutil.RemoveFinalizer(cr, dashboardFinalizer)
```
pb82, line 183: I think we need to discuss the delete flow. Currently what happens is that when you delete a dashboard, the finalizer never gets removed, because of https://github.com/grafana-operator/grafana-operator-experimental/pull/13/files#r934929966.

What happens is, that in the next cycle, because both finalizer and delete timestamp are still present, the delete flow is reached again. But this time deleting the dashboard from the instance fails because is has already been deleted. And then the finalizer is removed.

Is this the intended delete flow?

pb82:
@HubertStefanski great work, thanks! I have verified that creating and deleting dashboards works and it's great to have the Grafana client integrated now.

Just a few thoughts about the finalizer:

I agree with @NissesSenap that we should try to avoid them. It makes the Grafana Operator hard to integrate into other products and operators as they now need to contain logic to cleanup Grafana resources in the right order. The Api Management product is one such example.

Unfortunately, multi namespace support makes this even harder. If the Grafana Operator was constricted to a single namespace we could use ownership and `blockOwnerDeletion` to enforce cleanup order. But this is not allowed across namespaces. See [here](https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/) for an explanation why that is.

I'll try to come up with ideas about how to avoid finalizers and will write them in another comment.

pb82:
How we could avoid finalizers:

The problem is that when a dashboard is deleted, we don't know any longer to which Grafana instances it was added. This is because that information is part of the dashboard status (UIDforInstance).

So the key to avoid the finalizer is to move this information somewhere else. We probably don't want to add it to the Grafana status, because then you'd have the dashboard controller modify the Grafana CR.

Idea:

1. we keep the dashboard/instance relationships in a config map per Grafana CR (same namespace as the Grafana CR)
2. this config map lists all dashboards of that instance in a json structure `<namespace>: [{<name>: <UID>}]`
3. the config map gets updated by the dashboard controller. When a new dashboard is imported, a new entry is made.
4. when a dashboard gets deleted, the dashboard controller receives the name and the namespace (of the deleted dashboard CR).
5. the dashboard controller now checks *all* of those config maps (we can label them or add a reference to them in the Grafana CR).
6. if the dashboard is found in one of the maps (by namespace/name), it will be deleted from the respective instance via the UID.
7. the entry is removed from the config map and the config map is updated.

This comes with it's own problems:

1. if someone deletes this config map, deletion would on longer work. We also need some sort of reconciler to sync the config map with all present dashboards.

controllers/client/grafana_client.go
HubertStefanski: I think we can try to contribute this to the upstream library, I always prefer things to be working out-of-the-box, rather than having to do workarounds ourselves

HubertStefanski:
Closing, the changes have been merged outside this PR, as part of @pb82 work

PR #25, opened 2022-09-16
go-vet fix, reformatting, golang 1.19 

weisdd:
- Dockerfile:
  - Bumped golang version to 1.19;
  - Removed references to OS, so the same Dockerfile can be used for M1;
- Fixed a go vet complaint;
- Reformatted code with gofumpt.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Build the manager binary
-FROM golang:1.16 as builder
+FROM golang:1.19 as builder
 
 WORKDIR /workspace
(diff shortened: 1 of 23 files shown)


HubertStefanski (approved):
LGTM, I forgot to update to 1.19 thanks for the reminder!

weisdd:
I've also adjusted Dockerfile a moment ago - it was previously enforcing builds for amd64, thus could not be used on M1. I think, for now, it should rather rely on build-machine architecture.

PR #26, opened 2022-09-20
feat: reduce the number of kube api requests

pb82:
Reduced the number of Kuberbetes API requests by not reconciling Grafana CRs continuously (every 10 seconds). I think this is unnecessary, changes to the CR will be detected anyway.

Also expose metrics about the Grafana API requests.


HubertStefanski (approved):
LGTM!

PR #27, opened 2022-09-23
fix: nil pointer dereference, incorrect url for svc, and other fixes

weisdd:
- With `preferIngress: false`, the operator couldn't reach grafana instance, because the former was generating an incorrect URL (missing protocol). Example of the error:
`grafana-operator-exp-controller-manager-6575764768-gn688 manager 1.6639349543764849e+09 ERROR   controller.grafanadashboard     error reconciling dashboard  {"reconciler group": "grafana.integreatly.org", "reconciler kind": "GrafanaDashboard", "name": "grafanadashboard-sample-3", "namespace": "grafana-operator-experimental-system", "dashboard": "grafanadashboard-sample-3", "grafana": "grafana-a", "error": "Post \"grafana-a-service.grafana-operator-experimental-system.svc.cluster.local:3000\": unsupported protocol scheme \"grafana-a-service.grafana-operator-experimental-system.svc.cluster.local\""}`'
- Also, when the operator is unable to reach grafana instance, resp is nil, so there's nil pointer dereference in `nc (in *instrumentedRoundTripper) RoundTrip`, fixed that;
- Shortened `namePrefix` in kustomize, because service name was exceeding name length limits;
- Fixed example for Datasource + added support for `editable` field:
  - Though, the field doesn't actually impact datasource. Even when I tried to submit a datasource via curl, grafana would always respond with `readOnly: false`, so I guess it's not supposed to work through API or it's a bug in grafana. So, maybe it should have rather been completely removed from the operator;
- Changed `imagePullPolicy` to simplify development (I push images under the same tag).
- Bumped version of grafana image from `9.0.0` to `9.1.6` (latest).

### DIFF
--- a/controllers/client/grafana_client.go
+++ b/controllers/client/grafana_client.go
@@ -4,11 +4,12 @@ import (
 	"context"
 	"crypto/tls"
 	"errors"
-	"github.com/grafana-operator/grafana-operator-experimental/controllers/metrics"
 	"net/http"
 	"net/url"
 	"strconv"
 	"time"
+
+	"github.com/grafana-operator/grafana-operator-experimental/controllers/metrics"
 
 	"github.com/grafana-operator/grafana-operator-experimental/api/v1beta1"
 	"github.com/grafana-operator/grafana-operator-experimental/controllers/config"
@@ -37,12 +38,14 @@ func newInstrumentedRoundTripper(instanceName string) http.RoundTripper {
 
 func (in *instrumentedRoundTripper) RoundTrip(r *http.Request) (*http.Response, error) {
 	resp, err := in.wrapped.RoundTrip(r)
-	metrics.GrafanaApiRequests.WithLabelValues(
-		in.instanceName,
-		r.URL.Path,
-		r.Method,
-		strconv.Itoa(resp.StatusCode)).
-		Inc()
+	if resp != nil {
+		metrics.GrafanaApiRequests.WithLabelValues(
+			in.instanceName,
+			r.URL.Path,
+			r.Method,
+			strconv.Itoa(resp.StatusCode)).
+			Inc()
+	}
 	return resp, err
 }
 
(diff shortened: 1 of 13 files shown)


controllers/client/grafana_client.go
```diff
@@ -37,12 +38,14 @@ func newInstrumentedRoundTripper(instanceName string) http.RoundTripper {
 
 func (in *instrumentedRoundTripper) RoundTrip(r *http.Request) (*http.Response, error) {
 	resp, err := in.wrapped.RoundTrip(r)
-	metrics.GrafanaApiRequests.WithLabelValues(
-		in.instanceName,
-		r.URL.Path,
-		r.Method,
-		strconv.Itoa(resp.StatusCode)).
-		Inc()
+	if resp != nil {
```
pb82, line 41: Thanks, I noticed the same and meant to send a PR for it :+1:

controllers/reconcilers/grafana/grafana_service_reconciler.go
```diff
@@ -48,7 +48,12 @@ func (r *ServiceReconciler) Reconcile(ctx context.Context, cr *v1beta1.Grafana,
 
 	// try to assign the admin url
 	if !cr.PreferIngress() {
-		status.AdminUrl = fmt.Sprintf("%v.%v.svc.cluster.local:%d", service.Name, cr.Namespace,
+		protocol := "http"
+		if cr.Spec.Config["server"] != nil && cr.Spec.Config["server"]["protocol"] != "" {
```
pb82, line 52: can `cr.Spec.Config` be nil?

weisdd: I think it shouldn't be possible, because CRs cannot be deployed without `config` field:
`The Grafana "grafana-a" is invalid: spec.config: Required value`
But if you feel like it's needed, I could certainly add it.

Btw, discovered that CRs like this lead to operator crash:
```
---
kind: Grafana
metadata:
  name: grafana-a
  labels:
    dashboards: "grafana-a"
spec:
```

```
---
kind: Grafana
metadata:
  name: grafana-a
  labels:
    dashboards: "grafana-a"
```

weisdd: Added nil check for cr.Spec.Config, moved the logic to a helper function and added tests for it.

pb82 (approved):

Issue #20, opened 2022-09-07
Optimize interactions with kube-apiserver


weisdd:
Previously, some users complained about grafana-operator making too many requests to kube-apiserver (e.g. https://github.com/grafana-operator/grafana-operator/issues/460).
It'd be useful to:

- [ ] have a way to track such interactions (e.g. through metrics);
- [ ] optimize interactions if needed (use `WATCH` + cache results).

weisdd:
I'll try to work on this in the coming weeks. Since it's not the top priority, I think it should fit my schedule.

addreas:
This is the cache package thats in the controller-runtime: https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.13.0/pkg/cache

weisdd:
Before https://github.com/grafana-operator/grafana-operator-experimental/pull/26 was merged, we could see something like this (unfortunately, stats are not filtered by service account, so some of the entries are likely to come not from the operator):

```
 180 "get - /api/v1/namespaces/grafana-operator-experimental-system/configmaps/f75f3bba.integreatly.org"
 360 "get - /apis/coordination.k8s.io/v1/namespaces/grafana-operator-experimental-system/leases/f75f3bba.integreatly.org"
  13 "list - /api/v1/namespaces/grafana-operator-experimental-system/resourcequotas"
   1 "list - /api/v1/namespaces/grafana-operator-experimental-system/secrets?limit=500"
 180 "update - /api/v1/namespaces/grafana-operator-experimental-system/configmaps/f75f3bba.integreatly.org"
  37 "update - /api/v1/namespaces/grafana-operator-experimental-system/services/grafana-a-service"
  37 "update - /apis/apps/v1/namespaces/grafana-operator-experimental-system/deployments/grafana-a-deployment"
 180 "update - /apis/coordination.k8s.io/v1/namespaces/grafana-operator-experimental-system/leases/f75f3bba.integreatly.org"
   1 "watch - /api/v1/namespaces/grafana-operator-experimental-system/configmaps?allowWatchBookmarks=true&fieldSelector=metadata.name%3Dgrafana-a-ini&resourceVersion=1913&timeout=9m46s&timeoutSeconds=586&watch=true"
[...]

PR #28, opened 2022-09-23
feature: add audit lab

weisdd:
A simple lab that helps to keep track of number of API requests sent to kube-apiserver.

Things that we can potentially decide to change now or at some point later in time:

- Cluster name (`kind`) by default. I don't have multiple cluster at a time, so the default name is just fine for me. If that's not your case, feel free to suggest something else;
- Once RBAC issues are resolved (currently, kubebuilder generates roles with insufficient permissions), we can integrate the lab into Makefile (or something much simpler to use like Taskfile);
- It can also be part of an acceptance workflow if needed.

In support of https://github.com/grafana-operator/grafana-operator-experimental/issues/20

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,6 @@ vendor
 *~
 
 .vscode/
+
+labs/audit/kube-apiserver-audit.log
+.DS_Store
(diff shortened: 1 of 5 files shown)


pb82:
This looks great! Let's talk about it on today's call.

pb82:
Happy to merge this as a way to verify API access in the future.

PR #30, opened 2022-09-27
Namespaced mode

pb82:
Adds a `--namespace` flag to the Operator that allows to limit it's operation to a single namespace.


pb82:
Merging this as the `namespace` flag by itself can already be useful.
We still need to provide an example for using Operator Groups to manage permissions and find a way to  limit permissions when installing in namespaced mode from OLM.

PR #32, opened 2022-10-23
Examples

pb82:
Adds examples for providing external admin credentials and also updates the code to lookup the credentials so that the Operator can discover it.

Contains the following examples:

* Basic
* Credentials from config
* Credentials from secret
* OAuth Proxy

ping @HubertStefanski @NissesSenap @weisdd


api/v1beta1/typeoverrides.go
```diff
@@ -58,7 +58,7 @@ type DeploymentV1Spec struct {
 
 	Selector *metav1.LabelSelector `json:"selector,omitempty" protobuf:"bytes,2,opt,name=selector"`
 
-	Template *v14.PodTemplateSpec `json:"template,omitempty" protobuf:"bytes,3,opt,name=template"`
+	Template *DeploymentV1PodTemplateSpec `json:"template,omitempty" protobuf:"bytes,3,opt,name=template"`
```
pb82, line 61: This was required so we can override certain fields of the pod template without providing others.

examples/basic/resources.yaml
```diff
@@ -0,0 +1,54 @@
...
+        "list": []
+      },
+      "annotations": {
+        "list": []
+      },
+      "refresh": "5s",
+      "schemaVersion": 17,
+      "version": 0,
+      "links": []
+    }
```
weisdd, line 54: README.md says: "A basic deployment of Grafana with a dashboard and a datasource.", but the datasource is missing. So, either README.md or the example should be fixed :)

weisdd (changes requested):

controllers/client/grafana_client.go
```diff
@@ -2,51 +2,97 @@ package client
...
+			Namespace: grafana.Namespace,
+		}
+		err := c.Get(ctx, selector, secret)
+
+		if err != nil {
+			return nil, err
+		}
+
+		if secret.Data == nil {
+			return nil, nil
```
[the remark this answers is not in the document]
pb82: Yes, good point

controllers/metrics/metrics.go
```diff
@@ -26,10 +26,18 @@ var (
 		Name:      "requests",
 		Help:      "requests against the grafana api per instance",
 	}, []string{"instance_name", "path", "method", "status"})
+
+	DashboardUrlRequests = prometheus.NewCounterVec(prometheus.CounterOpts{
```
weisdd, line 30: Example of such metric: `grafana_operator_dashboards_requests{dashboard="grafanadashboard-from-url",method="GET",path="/integr8ly/grafana-operator/master/deploy/examples/remote/grafana-dashboard.json",status="200"} 2`
As you can see, there's no reference for a namespace, so I wonder what would happen when you have dashboards with the same name in multiple namespaces?

pb82: Good point. I wonder, do we even need to include the dashboard? The metric is meant to count the requests against the dashboard API overall, it doesn't have to be per dashboard.

weisdd: I think we're talking about different metrics here. My point was only about counting the number of requests for dashboards pulled from URLs. :)

pb82: Ah I see. This metric does count the number of requests against the Grafana dashboard API. I've now added the namespace to the dashboard name label.

pb82:
@weisdd are you ok with merging this?

weisdd:
@pb82 yep, let's merge it, and thanks for the fixes :)

PR #37, opened 2022-11-08
wip: initial dashboard sync on operator restart

pb82:
This PR contains the following changes:

1. Use the status block instead of annotations to store known dashboards and data sources
2. An initial dashboard sync-the-world to detect dashboards in Grafana that no longer have a CR representing them

The idea is that we only need to do this once after the Operator has (re)started. When the operator is running, it should not get into the situation where a dashboard CR gets deleted, but the dashboard is still present in Grafana. It could still happen if the Grafana API request fails, so there might be an reason to do this not only on startup. What do you think?

TODO: initial datasources sync, initial folder sync


pb82:
@weisdd based on our discussion last call, i've included additional metrics for the dashboard sync and started benchmarking it. We can also minimize the updates to Grafana CRs to only one update after all dashboards of this CR have been synced. I can talk a bit about the results in the next call.

weisdd:
At first, I thought it would make sense to merge `ResyncPeriod` and `ContentCacheDuration` (btw, the latter can be renamed if you have better suggestions, I just copied it from the old version of the operator), but now I'm not that sure. It's something we should think about.

PR #39, opened 2022-11-15
Design/Managing External Grafana Instance

elamaran11:
Hi @NissesSenap Per our discussion, have created a design document for a proposal for `Amazon Managed Grafana Integration` . Appreciate if you can take a look and let me know what you think. We can jump in to a call if needed.


documentation/Amazon-Managed-Grafana/ReadMe.md
```diff
@@ -0,0 +1,149 @@
...
+
+In order to accomplish this, "Reading keys for accessing Amazon Managed Grafana from AWS Secrets Manager" feature above becomes a prerequisite. `GrafanaDashboard` CRD should be modified to use `grafana-secret` to create `GrafanaDashboard` on Amazon Managed Grafana. This allows customers to use GitOps approach to create Grafana Dashboards on remote Grafana instances such as Amazon Managed Grafana.
+
+### Creating Grafana alerting on Amazon Managed Grafana.
+
+Today Grafana operator supports the creation of Grafana notification on self managed grafana instance in the cluster where Grafana operator is installed. Grafana operator should be enhanced to support creating notifications on Amazon Managed Grafana. 
+
+In order to accomplish this, "Reading keys for accessing Amazon Managed Grafana from AWS Secrets Manager" feature above becomes a prerequisite. `GrafanaNotificationChannel` CRD should be modified to use `grafana-secret` to create `GrafanaNotificationChannel` on Amazon Managed Grafana. This allows customers to use GitOps approach to create Grafana alerting on remote Grafana instances such as Amazon Managed Grafana.
```
NissesSenap, line 139: NotificationChannel is out of support in grafana 9 and the new version of the operator we will only support grafana 9 and forward.

documentation/Amazon-Managed-Grafana/ReadMe.md
```diff
@@ -0,0 +1,149 @@
...
+In short the proposal in this document is about enhancing the Grafana Operator to support the integration to managed grafana services such as Amazon Managed Grafana. We would need to enhance the current version of Grafana Operator to support the following :
+
+- Reading keys for accessing Amazon Managed Grafana from AWS Secrets Manager.
+- Adding AWS data sources to Amazon Managed Grafana.
+- Creating Grafana dashboards on Amazon Managed Grafana.
+- Creating alerting on Amazon Managed Grafana.
+
+### Reading keys for accessing Amazon Managed Grafana from AWS Secrets Manager
+
+Today Grafana operator supports Kubernetes role, rolebinding, service account mechanisms to access other kubernetes resources. We need to add a mechanism to allow Grafana Operator to read API keys for remote Grafana Instance from a secret store such as AWS Secrets Manager using [external-secrets](https://github.com/external-secrets/external-secrets).
```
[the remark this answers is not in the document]
NissesSenap: What would be interesting from an operator point of view is if we want to add IAM/IRSA support to the operator and thus being able to talk to the amazon managed grafana instance by setting well defined IAM rules.

NissesSenap (commented):
So I think we misunderstood each other, to me this is a long list of issues instead of a design proposal.

We need to have a basic discussion on how to design the CRD:s to support external grafana instances. Assuming that the secrets needed to connect to your external grafana instance is available inside the cluster how would we use these secrets to talk to the external grafana instance?

How would the grafana crd look like?
How would the datasource crd look like?

If you don't feel comfortable with go that is fine, we will figure out the controller part but as an end user how would you like the operator work with external resources?

documentation/Amazon-Managed-Grafana/ReadMe.md
```diff
@@ -0,0 +1,149 @@
...
+    kind: ClusterSecretStore
+  target:
+    name: grafana-secret # Secret name in k8s
+  data:
+  - secretKey: grafana-token # which key it's going to be stored
+    remoteRef:
+      key: grafana-api-key # Our secret-name goes here
+```
+
+### Adding AWS data sources to Amazon Managed Grafana
```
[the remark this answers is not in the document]
pb82: That's correct, we don't make any assumptions about the type of data source. You might have to install a plugin for a certain data source, but other than that, the datasource CR is compatible with all of them.

documentation/Amazon-Managed-Grafana/ReadMe.md
elamaran11: Sounds good, We need a mechanism to support alerting on external grafana instances.

documentation/Amazon-Managed-Grafana/ReadMe.md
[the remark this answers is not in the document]
elamaran11: Hi @NissesSenap I agree to your secrets manager comment. We dont have a need to read external secrets, i was callit as a nice to have feature but i agree taht secret should in cluster so the operator can use the same to access external grafana instance. 
With respect to IRSA, This is something we can discuss but im not sure if i get around the need for IRSA. The recommended approach we have seen is using Grafana Keys and use Grafana APIs to perform operations on AMG.

documentation/Amazon-Managed-Grafana/ReadMe.md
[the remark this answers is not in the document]
elamaran11: Understood @pb82 @NissesSenap Thankyou for clarifying this. We may need to add couple of samples on using other AWS data sources in the `examples` folder.

documentation/Amazon-Managed-Grafana/ReadMe.md
```diff
@@ -0,0 +1,149 @@
...
+
+In order to accomplish this, "Reading keys for accessing Amazon Managed Grafana from AWS Secrets Manager" feature above becomes a prerequisite. `GrafanaDataSource` CRD should be modified to use `grafana-secret` to add AWS data sources as `GrafanaDataSource` on Amazon Managed Grafana. This allows customers to use GitOps approach to add AWS data sources as `GrafanaDataSource` on remote Grafana instances such as Amazon Managed Grafana.
+
+### Creating Grafana dashboards on Amazon Managed Grafana.
+
+Today Grafana operator supports the creation of Grafana dashboards on self managed grafana instance in the cluster where the Grafana operator is installed. Grafana operator should be enhanced to support creating dashboards on Amazon Managed Grafana. 
+
+In order to accomplish this, "Reading keys for accessing Amazon Managed Grafana from AWS Secrets Manager" feature above becomes a prerequisite. `GrafanaDashboard` CRD should be modified to use `grafana-secret` to create `GrafanaDashboard` on Amazon Managed Grafana. This allows customers to use GitOps approach to create Grafana Dashboards on remote Grafana instances such as Amazon Managed Grafana.
```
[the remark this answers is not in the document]
elamaran11: > So I think we misunderstood each other, to me this is a long list of issues instead of a design proposal.
> 
> We need to have a basic discussion on how to design the CRD:s to support external grafana instances. Assuming that the secrets needed to connect to your external grafana instance is available inside the cluster how would we use these secrets to talk to the external grafana instance?
> 
> How would the grafana crd look like? How would the datasource crd look like?
> 
> If you don't feel comfortable with go that is fine, we will figure out the controller part but as an end user how would you like the operator work with external resources?

Thankyou for responding @NissesSenap We can definately jump in to a call to discuss on open items and next steps and we can also discuss over slack. From and enduser perspective, the operator should be able to use the secrets in the cluster and access external grafana instance to add data sources like AMP, CW and also create Dashboards and alerting on remote AMG using same or new CRDs.

documentation/Amazon-Managed-Grafana/ReadMe.md
NissesSenap: It's a separate CRD and it's out of scope for this design.
See: https://github.com/grafana-operator/grafana-operator-experimental/issues/38

documentation/Amazon-Managed-Grafana/ReadMe.md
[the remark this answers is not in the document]
NissesSenap: We are happy to accept a PR around it

documentation/Amazon-Managed-Grafana/ReadMe.md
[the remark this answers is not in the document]
NissesSenap: Okay that is good to know.

elamaran11:
> I think most pieces are already there to support externally managed Grafana instances. We just need a way to tell the Operator than a given Grafana CR refers to an external, already existing instance and where the credentials are located. We could do this by adding new configuration options to the Grafana CR spec, e.g.
> 
> ```
> spec:
>   external:
>     url: <external grafana url, type string>
>     admin_username: <type SecretKeySelector>
>     admin_password: <type SecretKeySelector>
> ```
> 
> all other fields in the spec would be ignored for external instances. The url gets reflected into the status, so the dashboad and datasource reconcilers know how to talk to those instances. We'll have to add some logic to the Grafana client wrapper to obtain the credentials from the secret key selectors (instead of the generated secret).

@pb82 The approach looks reasonable and good but only caveat is each and every dashboard will have dupllication of snippets for URL and API reference for external grafana. Im working on a different approach for this. I will submit an updated Pr.

pb82:
Thanks @elamaran11 , looking forward to your proposal!

elamaran11:
Hi @pb82 @NissesSenap Please check this updated design proposal. Please let me know what you think. Happy to jump in to a call.

documentation/External-Grafana/ReadMe.md
```diff
@@ -0,0 +1,179 @@
...
+    type: cloudwatch
+    region: <type ConfigMapSelector>
+    iamRole: <type ConfigMapSelector>
+```
+
+Note: `grafanainstance` is not mandatory if its setup as default external Grafana instance in `GrafanaInstance` CRD. 
+
+This is not a road block feature but nice to have to support all cloud based data sources.
+
+### Creating Grafana alerting on Amazon Managed Grafana.
```
NissesSenap, line 139: To get smaller scopes I would really like to have GrafanaAlerting CRD in a separate PR.
It will make it easier to discuss how we should solve things and also easier to close specific tasks.

documentation/External-Grafana/ReadMe.md
```diff
@@ -0,0 +1,179 @@
...
+## Proposal
+
+In short the proposal in this document is about enhancing the Grafana Operator to support the integration to managed grafana services. We would need to enhance the current version of Grafana Operator to support the following :
+
+- Defining external Grafana instance as Grafana source to Grafana operator.
+- Creating Grafana dashboards on external Grafana instance.
+- Adding data sources to external Grafana instance.
+- Creating alerting on external Grafana instance.
+
+### Defining external Grafana instance as Grafana source to Grafana operator.
```
NissesSenap, line 52: We don't think we need a new CRD, we should be able to use the existing one.
I have provided an example below.

documentation/External-Grafana/ReadMe.md
```diff
@@ -0,0 +1,179 @@
...
+
+Today Grafana operator supports self managed grafana instance as a Grafana source to Grafana operator. With this new feature, we would like to add a new CRD by name `GrafanaInstance` which basically creates an identity to an external Grafana instance. 
+
+> CRD design of how a `GrafanaInstance` will look like. In this sample design, `url` should be preloaded in a ConfigMap and `grafana_api_key` should be loaded to a Secret. Choice of using external secrets or loading secrets manually is end user responsibility :
+
+```.yaml
```
NissesSenap, line 58: Example using the existing grafana CRD to talk to an external grafana instance.

```.yaml
kind: Grafana
metadata:
  annotations:
  labels:
    app: grafana
  name: grafana
  namespace: grafana-operator-system
spec:
  external:
    url: <external grafana url, type string>
    admin_username: <type SecretKeySelector>
    admin_password: <type SecretKeySelector>
```

If external is defined we won't create any new instance and instead try to talk to the external one.
We will probably have to change some CRD definitions in the code and make some features optional but that should be okay.

documentation/External-Grafana/ReadMe.md
```diff
@@ -0,0 +1,179 @@
...
+
+Currently Grafana operator has an integration to add Amazon Managed Service for Promethus (AMP) as a data source to Grafana instances hosted in a kubernetes environment. As more customers starting to use external grafana services, expanding the Grafana operator to support remote Grafana instances becomes inevitable. Adding ability to integrate with external grafana services, adding data sources, creating dashboards and alerting on a remote Grafana instances offloads responsibilities of managing a grafana instanaces from ops personas which helps them to focus on developing the features required for their business. This helps the customer teams to move from self managed Grafana instance on their Kubernetes environments to Pay as you go model on Grafana instances provided by providers.
+
+## Out of scope
+
+In the design, integration to remote managed Grafana providers other than Amazon Managed Grafana is out of scope at this point. We may add additional integrations as part of future road map items.
```
NissesSenap, line 26: This can be removed

documentation/External-Grafana/ReadMe.md
```diff
@@ -0,0 +1,179 @@
...
+
+## Verification
```
NissesSenap, line 28: This is specific to amazon and it don't really matter, there is no reason why this design document can't solve all external grafana configs.

So the only thing is
`- Create integration tests to create alerting on a remote Grafana Instance.`

documentation/External-Grafana/ReadMe.md
```diff
@@ -0,0 +1,179 @@
...
+  name: amazon-managed-grafana
+  annotations:
+    grafanainstance.kubernetes.io/is-default-grafana: "true"
+spec:
+  external:
+    url: <type ConfigMapSelector>
+    grafana_api_key: <type SecretKeySelector>
+```
+
+### Creating Grafana dashboards on Amazon Managed Grafana.
```
NissesSenap, line 84: This will be solved due to the label selectors that we already have in GrafanaDashboard so this can be removed.
We will solve this with the standard `Grafana` kind as mentioned earlier.

documentation/External-Grafana/ReadMe.md
```diff
@@ -0,0 +1,179 @@
...
+metadata: 
+  name: grafana-dashboard-from-url
+  labels: 
+    app: grafana
+spec:
+  grafanainstance: "amazon-managed-grafana"
+  url: https://raw.githubusercontent.com/integr8ly/grafana-operator/master/deploy/examples/remote/grafana-dashboard.json
+```
+
+### Adding data sources to external Grafana instance.
```
NissesSenap, line 102: This will be solved due to the label selectors that we already have in GrafanaDatasource so this can be removed.
We will solve this with the standard Grafana kind as mentioned earlier.

NissesSenap (commented):
I understand that my end @pb82 comments was a bit unclear but we think that we can use the existing `Grafana` kind so no neeed to create a new CRD/controller.

So you can remove that part.
About alerting please write a separate proposal around it since it's a new CRD and there we will need to look around much more at the Alerting API.

Since this proposal will solve all external grafana instances you can remove most of the Amazon specific mentions but I do like the example of pointing to a good use-case.

documentation/Amazon-Managed-Grafana/ReadMe.md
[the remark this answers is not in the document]
elamaran11: Sure @NissesSenap Once we have external Grafana Instance working, i can submit a PR.

documentation/External-Grafana/ReadMe.md
elamaran11: @NissesSenap I will work on a new Design doc PR on this as soon as possible. I agree we can decouple alerting from this PR at this point of time.

documentation/External-Grafana/ReadMe.md
elamaran11: I have made this change.

documentation/External-Grafana/ReadMe.md
elamaran11: Done.

documentation/External-Grafana/ReadMe.md
elamaran11: Done.

documentation/External-Grafana/ReadMe.md
elamaran11: Agreed and fixed.

elamaran11:
> I understand that my end @pb82 comments was a bit unclear but we think that we can use the existing `Grafana` kind so no neeed to create a new CRD/controller.
> 
> So you can remove that part. About alerting please write a separate proposal around it since it's a new CRD and there we will need to look around much more at the Alerting API.
> 
> Since this proposal will solve all external grafana instances you can remove most of the Amazon specific mentions but I do like the example of pointing to a good use-case.

I have fixed all comments on the PR. I also agree to the approach of using existing `Grafana` CRD which simplies the process even more. Only thing i want to iterate is, We want to use `url` and `grafana_api_key` as an approach to access remote Grafana Instance such as AMG to use the same while adding data souces and while creating dashboards.

documentation/External-Grafana/ReadMe.md
```diff
@@ -0,0 +1,68 @@
...
+kind: Grafana
+metadata:
+  annotations:
+  labels:
+    app: grafana
+  name: grafana
+  namespace: grafana-operator-system
+spec:
+  external:
+    url: <type ConfigMapSelector>
```
NissesSenap, line 60: Hmm, why do you think it needs to be a ConfigMapSelector?
Isn't it enough with a string to the specific URL just as @pb82 described?

``` 
    url: <external grafana url, type string>
```

documentation/External-Grafana/ReadMe.md
```diff
@@ -0,0 +1,68 @@
...
+    grafana_api_key: <type SecretKeySelector>
```
NissesSenap, line 61: Hmm you have a good point about using grafana_api_key.
First of all I think api_key is a better variable name. I at least think assumed since we are in the grafana instance config.

But the second question is how do we want to login to external grafana instances.
A best practice is decadently to use a API key instead of doing

``` 
    admin_username: <type SecretKeySelector>
    admin_password: <type SecretKeySelector>
``` 

The question is if we want to support the other possible solution as well?
Or we can just say that any external grafana instance should always be using a API key.
Or we could support both by using `username` & `password`.
And if you are using an API key you could set username as empty or something like that, i have seen similar solutions on a few API:s through out the years.

But it's not super straight forward and we would have to do allot of docs.
What do you think @elamaran11  and @pb82 ?

NissesSenap (commented):
Added some minor comments.

documentation/External-Grafana/ReadMe.md
elamaran11: I think `ConfigMapSelector` is not needed. We can go with using url pointed directly as above. I fixed the PR.

documentation/External-Grafana/ReadMe.md
elamaran11: HI @NissesSenap IMO we should support both authentication mechanisms. `API_KEY` and `username`, `password` so that the user has a flexbility to choose the same accordingly to their needs. We should support separate paremeters for `API_KEY` and `username`, `password`. If `API_KEY` is passed that should take precedence over `username`, `password`.

NissesSenap: Yeah it's probably easier then doing some strange workaround supporting two in one.

Please describe all of the three options in the document and write somewhere that API key should take precedence over username, password. 
Then we don't forget this during the implementation :)

elamaran11: @NissesSenap Added, Please check. I have covered all 3 variants of options.

pb82: Yes, agree to support both. So basically:

```
spec:
  external:
      admin_username: <type SecretKeySelector>
      admin_password: <type SecretKeySelector>
      admin_api_key:  <type SecretKeySelector>
```

right? And if the API key is provided, the othe two fields are ignored (if provided).

elamaran11: @pb82 Thats right. If API Key is provided, that takes precedence over U/P. But also remember `url` should also a mandatory field for external grafana instance.

pb82: @elamaran11 right, I forgot to include the url.

NissesSenap (approved):
LGTM, do you have any other thoughts @pb82 ?

pb82 (approved):

PR #42, opened 2022-11-21
feat: dashboards caching, fix: detect changes in dashboards with no .Spec.Json

weisdd:
- fix: detect changes in dashboards with no .Spec.Json
  - `Spec.Json` is used to calculate hash and, thus to detect changes in dashboards, but not every type of dashboard has the field defined. So, we should keep the field updated before calculating hash to make sure changes in dashboards get noticed by the controller;
- feat: added support for URL-based dashboards caching.
 
NOTE: As discussed during one of our meetings, we don't currently have a periodic reconciliation for dashboards. Due to that, with caching added, the only way to refresh a URL-based dashboard is to change a URL. We should implement a mechanism to queue URL-based dashboards reconciliations. I think you had an idea on how to do that @pb82

Fixes: #33


pb82:
Verified this, along with the resyncPeriod. I think it's actually ok to have both, they do different things:

* resyncPeriod: determines how often a dashboard will be refreshed in Grafana, regardless of  content changes
* contentCacheDuration: only applies to dashboards from urls, determines how long the content is not re-fetched from the url

pb82 (approved):

PR #43, opened 2022-12-04
Add support to configure external grafana

NissesSenap:
This PR implements support to configure external grafana instance as designed in: https://github.com/grafana-operator/grafana-operator-experimental/pull/39

Have written a simple example using the grafana-operator that showcases how to use this feature.


pb82:
@NissesSenap this should be good to go now. Do you want to squash the commits?

NissesSenap:
FYI: @elamaran11

controllers/dashboard_controller.go
```diff
@@ -290,6 +294,10 @@ func (r *GrafanaDashboardReconciler) onDashboardCreated(ctx context.Context, gra
 		return err
 	}
 
+	if grafana.Spec.External != nil && cr.Spec.Plugins != nil {
```
NissesSenap, line 297: Potential controversial check

NissesSenap (commented):
Thanks for the help @pb82 , I have rebased and done some testing on my setup.
I have also added a few examples.

I have also added a potentially controversial check

If a grafanadashboard/datasource matches with a grafana instances that is external the operator will throw an error and it will show in the logs.

Since we can't apply any plugins to external instances I thought this would make sense. The bad thing with this solution is that we don't show anything on the status of the cr.
So it might be hard for the user to find this error.

There is a potential use-case where you might want a external grafana instance together with a local grafana instance use the same dashboard selector. Assuming that I would be willing to install the needed plugin manually to the external grafana instance.

But I think it would be very limited times this would happen, even if could support both solutions on the same operator you probably wouldn't. 
And if you for some reason did you probably won't have the same dashboards on both the setups.
And in general I think it's so easy to add config in kubernetes so the few times this might happen it's probably better to have duplicates grafanadashobards where the second dashboard don't got a plugin defined.

This way your grafana admin wouldn't be confused when the plugin version don't match vs the manually managed plugins vs twhat is configured in the dashboard cr.

Might be worth to add a status message about this on the dashboard and datasource resource though.

elamaran11:
> FYI: @elamaran11

Awesome, thanks for adding me to PR @NissesSenap

controllers/dashboard_controller.go
```diff
@@ -201,13 +201,15 @@ func (r *GrafanaDashboardReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
 		}
 
-		// first reconcile the plugins
-		// append the requested dashboards to a configmap from where the
-		// grafana reconciler will pick them up
-		err = ReconcilePlugins(ctx, r.Client, r.Scheme, &grafana, dashboard.Spec.Plugins, fmt.Sprintf("%v-dashboard", dashboard.Name))
-		if err != nil {
-			controllerLog.Error(err, "error reconciling plugins", "dashboard", dashboard.Name, "grafana", grafana.Name)
-			success = false
+		if grafana.Spec.External == nil {
```
weisdd, line 204: How do you feel about introducing a struct method for that instead?

```go
func (in *Grafana) IsInternal() bool {
	return in.Spec.External == nil
}
```

The same for external instances:
```go
func (in *Grafana) IsExternal() bool {
	return in.Spec.External != nil
}
```
or even through negation (Python protocols style):
```go
func (in *Grafana) IsExternal() bool {
	return !in.IsInternal()
}
```

Not like those are necessary, I just think it'll be easier to read the code. :)

NissesSenap: Agreed, I have updated to PR to match

weisdd (approved):
LGTM

Issue #41, opened 2022-11-17
Kubernetes https assumed on preferIngress


NissesSenap:
In line https://github.com/grafana-operator/grafana-operator-experimental/blob/7add20fdff6fc8ec60925b6cfc79a006f721b6c3/controllers/reconcilers/grafana/ingress_reconciler.go#L70 we assume that we are using `https` as the schema.

It's not a good idea but it's not impossible that the user uses a `http` endpoint and not `https`.
I guess we could look if `ingress.spec.tls != nil`, and if so we could assume that the user uses https. 

This is very low priority.

Issue #40, opened 2022-11-16
Kubernetes ingress missing url


NissesSenap:
I'm having issues trying to create grafana instance in kubernetes when using `make run`

I have created a simple grafana instance using the example files with some extra additions.

```.yaml
---
apiVersion: grafana.integreatly.org/v1beta1
kind: Grafana
metadata:
  name: grafana
  labels:
    dashboards: "grafana"
spec:
  client:
    preferIngress: true
  config:
    log:
      mode: "console"
    auth:
      disable_login_form: "false"
    security:
      admin_user: root
      admin_password: secret
  ingress:
    spec:
      ingressClassName: nginx
      rules:
        - host: test.io
          http:
            paths:
              - backend:
                  service:
                    name: grafana-service
                    port:
                      number: 3000
                path: /
                pathType: Prefix
---
apiVersion: grafana.integreatly.org/v1beta1
kind: GrafanaDashboard
metadata:
  name: grafana-dashboard
spec:
  instanceSelector:
    matchLabels:
      dashboards: "grafana"
  json: >
    {
      "id": null,
      "title": "Simple Dashboard",
      "tags": [],
      "style": "dark",
      "timezone": "browser",
      "editable": true,
      "hideControls": false,
      "graphTooltip": 1,
      "panels": [],
      "time": {
        "from": "now-6h",
        "to": "now"
      },
      "timepicker": {
        "time_options": [],
        "refresh_intervals": []
      },
      "templating": {
        "list": []
      },
      "annotations": {
        "list": []
      },
      "refresh": "5s",
      "schemaVersion": 17,
      "version": 0,
      "links": []
    }
---
apiVersion: grafana.integreatly.org/v1beta1
kind: GrafanaDatasource
metadata:
  name: example-grafanadatasource
spec:
  datasource:
    access: proxy
    database: prometheus
    jsonData:
      timeInterval: 5s
      tlsSkipVerify: true
    name: Prometheus
    url: http://prometheus-service:9090
  instanceSelector:
    matchLabels:
      dashboards: "grafana"
```

```.txt
[...]

PR #45, opened 2022-12-06
fix: properly generate adminURL for Ingresses, requeue grafana reconciliation

weisdd:
The PR contains multiple fixes for `adminURL` synchronisation:
- issue: operator can generate broken adminURLs in case IP / Host is missing (e.g. `http://`);
  - fix: always make sure hostname is set;
- issue: host is taken from `Status.LoadBalancer`, which will not contain the actual ingress host;
  - fix: use hosts instead;
- issue: operator always assumes https;
  - fix: if we have a matching host in tls settings, use https, for other cases - http;
- issue: if there's an error during Grafana reconciliation, the process gets stuck until there's any change in grafana spec;
  - fix: requeue grafana in case of errors;

I think it'll close the following issues:

Closes: #40 
Closes: #41

### DIFF
--- a/controllers/grafana_controller.go
+++ b/controllers/grafana_controller.go
@@ -102,9 +102,7 @@ func (r *GrafanaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		nextStatus.StageStatus = status
 
 		if status != grafanav1beta1.OperatorStageResultSuccess {
-			controllerLog.Info("stage in progress", "stage", stage)
-			finished = false
-			break
+			return ctrl.Result{RequeueAfter: RequeueDelayError}, nil
 		}
 	}
 
--- a/controllers/reconcilers/grafana/ingress_reconciler.go
+++ b/controllers/reconcilers/grafana/ingress_reconciler.go
@@ -67,7 +67,7 @@ func (r *IngressReconciler) reconcileIngress(ctx context.Context, cr *v1beta1.Gr
 		adminURL := r.getIngressAdminURL(ingress)
 
 		if len(ingress.Status.LoadBalancer.Ingress) == 0 {
-			return v1beta1.OperatorStageResultFailed, fmt.Errorf("ingress is not ready yet")
+			return v1beta1.OperatorStageResultInProgress, fmt.Errorf("ingress is not ready yet")
 		}
 
 		if adminURL == "" {
--- a/controllers/reconcilers/grafana/ingress_reconciler.go
+++ b/controllers/reconcilers/grafana/ingress_reconciler.go
@@ -64,13 +64,17 @@ func (r *IngressReconciler) reconcileIngress(ctx context.Context, cr *v1beta1.Gr
 
 	// try to assign the admin url
 	if cr.PreferIngress() {
-		if len(ingress.Status.LoadBalancer.Ingress) > 0 {
-			ingress := ingress.Status.LoadBalancer.Ingress[0]
-			if ingress.Hostname != "" {
-				status.AdminUrl = fmt.Sprintf("https://%v", ingress.Hostname)
-			}
-			status.AdminUrl = fmt.Sprintf("https://%v", ingress.IP)
+		adminURL := r.getIngressAdminURL(ingress)
+
+		if len(ingress.Status.LoadBalancer.Ingress) == 0 {
+			return v1beta1.OperatorStageResultFailed, fmt.Errorf("ingress is not ready yet")
 		}
+
+		if adminURL == "" {
+			return v1beta1.OperatorStageResultFailed, fmt.Errorf("ingress spec is incomplete")
+		}
+
+		status.AdminUrl = adminURL
 	}
 
 	return v1beta1.OperatorStageResultSuccess, nil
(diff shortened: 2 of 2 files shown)


pb82:
@weisdd do we require the user to provide an ingress host?

pb82:
The Operator will get stuck when the host is not provided. On k3d though, that's not needed. The Operator can use the IP address of the Ingress and that seems to work fine. I'll see if I can support that along your changes.

weisdd:
@pb82 I think that's always an expectation when using Ingresses, otherwise the traffic will get routed to a default backend. How does it work on k3d when you have multiple ingresses?

controllers/reconcilers/grafana/ingress_reconciler.go
```diff
@@ -64,13 +64,17 @@ func (r *IngressReconciler) reconcileIngress(ctx context.Context, cr *v1beta1.Gr
...
-		if len(ingress.Status.LoadBalancer.Ingress) > 0 {
-			ingress := ingress.Status.LoadBalancer.Ingress[0]
-			if ingress.Hostname != "" {
-				status.AdminUrl = fmt.Sprintf("https://%v", ingress.Hostname)
-			}
-			status.AdminUrl = fmt.Sprintf("https://%v", ingress.IP)
+		adminURL := r.getIngressAdminURL(ingress)
+
+		if len(ingress.Status.LoadBalancer.Ingress) == 0 {
+			return v1beta1.OperatorStageResultFailed, fmt.Errorf("ingress is not ready yet")
```
pb82, line 70: we can return `OperatorStageResultInProgress` here, it's not an error if the Ingress is not ready

pb82:
@weisdd I've added this to the getIngressAdminUrl function, which makes it work on k3d without providing a hostname:

```
	// if all fails, try to get access trough the load balancer
	var loadBalancerIp = ""
	for _, lb := range ingress.Status.LoadBalancer.Ingress {
		if lb.Hostname != "" {
			hostname = lb.Hostname
			break
		}
		if lb.IP != "" {
			loadBalancerIp = lb.IP
		}
	}

	if hostname == "" && loadBalancerIp != "" {
		hostname = loadBalancerIp
	}
```

what do you think? Is it ever reasonable to use the load balancer?

controllers/grafana_controller.go
```diff
@@ -102,9 +102,7 @@ func (r *GrafanaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		nextStatus.StageStatus = status
 
 		if status != grafanav1beta1.OperatorStageResultSuccess {
-			controllerLog.Info("stage in progress", "stage", stage)
-			finished = false
-			break
+			return ctrl.Result{RequeueAfter: RequeueDelayError}, nil
```
pb82, line 105: In that case the status would not get updated if a stage fails and the error message wouldn't be visible in the CR status.

pb82: why return  here immediately instead of letting the status update happen?

NissesSenap:
So I have verified that  below yaml is working as intended. The grafan adminURL reports `http://test.io`

```.yaml
---
kind: Grafana
metadata:
  name: grafana
  labels:
    dashboards: "grafana"
spec:
  client:
    preferIngress: true
  config:
    log:
      mode: "console"
    auth:
      disable_login_form: "false"
    security:
      admin_user: root
      admin_password: secret
  ingress:
    spec:
      ingressClassName: nginx
      rules:
        - host: test.io
          http:
            paths:
              - backend:
                  service:
                    name: grafana-service
                    port:
                      number: 3000
                path: /
                pathType: Prefix
```

controllers/grafana_controller.go
weisdd: @pb82 it's something that I briefly mentioned during the last meeting. Not like I checked that, but I think status update will immediately trigger another reconciliation attempt (due to change passed through a WATCH session) instead of letting the operator wait for the `RequeueDelayError` period. WDYT?

controllers/reconcilers/grafana/ingress_reconciler.go
weisdd: Definitely makes sense, fixed.

weisdd:
> 



> @weisdd I've added this to the getIngressAdminUrl function, which makes it work on k3d without providing a hostname:
> 
> ```
> 	// if all fails, try to get access trough the load balancer
> 	var loadBalancerIp = ""
> 	for _, lb := range ingress.Status.LoadBalancer.Ingress {
> 		if lb.Hostname != "" {
> 			hostname = lb.Hostname
> 			break
> 		}
> 		if lb.IP != "" {
> 			loadBalancerIp = lb.IP
> 		}
> 	}
> 
> 	if hostname == "" && loadBalancerIp != "" {
> 		hostname = loadBalancerIp
> 	}
> ```
> 
> what do you think? Is it ever reasonable to use the load balancer?

Well, single service ingresses are very unlikely to be used in "real" clusters.
In most setups, I think, there would be just a single Load-Balancer for an ingress-controller, and all Ingresses matching ingressClassName would be served through one IP.

In my local setup, I'm just using ingress-nginx and special DNS-names like: http://grafana.192.168.64.3.nip.io/ (it resolved in 192.168.64.3 - the IP I have for microk8s).

If you want to make it work for the setup you're describing, then yeah, we can make the change.

NissesSenap:
Did a rebase from master branch

NissesSenap:
I will merge this one. I know that you wanted to check something else @pb82 but I can't remember what.
But it's nothing that we can't have in a separate issue/PR.

PR #47, opened 2022-12-11
Use signal.NotifyContext for easier stop solution

NissesSenap:
Please ignore all the other changes, they are fixed by my editor.

The interesting part is that I deleted the stop channel and use signal.NotifyContext instead.

```.go
	stop := make(chan bool)
	sigs := make(chan os.Signal)
	signal.Notify(sigs, syscall.SIGTERM, syscall.SIGABRT, syscall.SIGINT)
	go func() {
		sig := <-sigs
		switch sig {
		case syscall.SIGTERM, syscall.SIGABRT, syscall.SIGINT:
			close(stop)
		}
	}()
```

And instead use:

```.go
	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt)
	defer stop()
```

### DIFF
--- a/controllers/grafanafolder_controller.go
+++ b/controllers/grafanafolder_controller.go
@@ -219,6 +219,8 @@ func (r *GrafanaFolderReconciler) SetupWithManager(mgr ctrl.Manager, ctx context
 		go func() {
 			for {
 				select {
+				case <-ctx.Done():
+					return
 				case <-time.After(d):
 					result, err := r.Reconcile(ctx, ctrl.Request{})
 					if err != nil {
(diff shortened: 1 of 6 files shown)


controllers/grafanafolder_controller.go
```diff
@@ -218,10 +219,8 @@ func (r *GrafanaFolderReconciler) SetupWithManager(mgr ctrl.Manager, stop chan b
 		go func() {
 			for {
 				select {
-				case <-stop:
```
simongottschlag, line 221: You should most likely listen for `<-ctx.Done()` now

NissesSenap: Didn't think I needed it due to ctx now including stop.
I re added it any way. 
Thanks for the check @simongottschlag .

Issue #7, opened 2022-05-30
[5.0] grafana CRD secure by default with more configuration options


NissesSenap:
From https://github.com/grafana-operator/grafana-operator/blob/master/documentation/roadmap/roadmap-v5.md
As part of 5.0 we would like to see the grafana-opreator to be secure by default + being more configurable.

We have done a design document that describes the new ideas in: https://github.com/grafana-operator/grafana-operator/blob/master/documentation/design_proposal/002_grafana_v1beta1.md

NissesSenap:
It's not perfect but we got a start. I will close this issue

PR #48, opened 2022-12-14
Adding deployment best practices to grafana instance

NissesSenap:
These settings should be used as default in kubernetes.
We will get issues with it on OCP though, due to how `RunAsUser` is managed and OCP automatically gives all containers a random user to run on per namespace/project.

In general we need a good way of detecting if we are running on OCP.
Currently the one we are using now is: https://github.com/grafana-operator/grafana-operator-experimental/blob/d0fcdf279f9aefbd8e2175593a63e00abd51119b/controllers/reconcilers/grafana/ingress_reconciler.go#L101

But it's a bit painful.

When it's done it should hopefully solve: https://github.com/grafana-operator/grafana-operator-experimental/issues/7

### DIFF
--- a/controllers/reconcilers/grafana/deployment_reconciler.go
+++ b/controllers/reconcilers/grafana/deployment_reconciler.go
@@ -13,38 +13,50 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	v13 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
 const (
-	InitMemoryRequest = "128Mi"
-	InitCpuRequest    = "250m"
-	InitMemoryLimit   = "512Mi"
-	InitCpuLimit      = "1000m"
-	MemoryRequest     = "256Mi"
-	CpuRequest        = "100m"
-	MemoryLimit       = "1024Mi"
-	CpuLimit          = "500m"
+	InitMemoryRequest                       = "128Mi"
+	InitCpuRequest                          = "250m"
+	InitMemoryLimit                         = "512Mi"
+	InitCpuLimit                            = "1000m"
+	MemoryRequest                           = "256Mi"
+	CpuRequest                              = "100m"
+	MemoryLimit                             = "1024Mi"
+	CpuLimit                                = "500m"
+	GrafanaHealthEndpoint                   = "/api/health"
+	ReadinessProbeFailureThreshold    int32 = 1
+	ReadinessProbeInitialDelaySeconds int32 = 5
+	ReadinessProbePeriodSeconds       int32 = 10
+	ReadinessProbeSuccessThreshold    int32 = 1
+	ReadinessProbeTimeoutSeconds      int32 = 3
 )
 
 type DeploymentReconciler struct {
-	client client.Client
+	client      client.Client
+	isOpenShift bool
 }
 
-func NewDeploymentReconciler(client client.Client) reconcilers.OperatorGrafanaReconciler {
+func NewDeploymentReconciler(client client.Client, isOpenShift bool) reconcilers.OperatorGrafanaReconciler {
 	return &DeploymentReconciler{
-		client: client,
+		client:      client,
+		isOpenShift: isOpenShift,
 	}
 }
 
 func (r *DeploymentReconciler) Reconcile(ctx context.Context, cr *v1beta1.Grafana, status *v1beta1.GrafanaStatus, vars *v1beta1.OperatorReconcileVars, scheme *runtime.Scheme) (v1beta1.OperatorStageStatus, error) {
-	_ = log.FromContext(ctx)
+	logger := log.FromContext(ctx)
+
+	openshiftPlatform := r.isOpenShift
+	logger.Info("reconciling deployment", "openshift", openshiftPlatform)
 
 	deployment := model.GetGrafanaDeployment(cr, scheme)
 	_, err := controllerutil.CreateOrUpdate(ctx, r.client, deployment, func() error {
-		deployment.Spec = getDeploymentSpec(cr, deployment.Name, scheme, vars)
+		deployment.Spec = getDeploymentSpec(cr, deployment.Name, scheme, vars, openshiftPlatform)
 		err := v1beta1.Merge(deployment, cr.Spec.Deployment)
 		return err
 	})
@@ -126,7 +138,7 @@ func getVolumeMounts(cr *v1beta1.Grafana, scheme *runtime.Scheme) []v1.VolumeMou
 	return mounts
 }
 
-func getContainers(cr *v1beta1.Grafana, scheme *runtime.Scheme, vars *v1beta1.OperatorReconcileVars) []v1.Container { // nolint
+func getContainers(cr *v1beta1.Grafana, scheme *runtime.Scheme, vars *v1beta1.OperatorReconcileVars, openshiftPlatform bool) []v1.Container { // nolint
 	var containers []v1.Container // nolint
 
 	image := fmt.Sprintf("%s:%s", config2.GrafanaImage, config2.GrafanaVersion)
@@ -178,6 +190,8 @@ func getContainers(cr *v1beta1.Grafana, scheme *runtime.Scheme, vars *v1beta1.Op
 		TerminationMessagePath:   "/dev/termination-log",
 		TerminationMessagePolicy: "File",
 		ImagePullPolicy:          "IfNotPresent",
+		SecurityContext:          getGrafanaContainerSecurityContext(openshiftPlatform),
+		ReadinessProbe:           getReadinessProbe(cr),
 	})
 
 	// Use auto generated admin account?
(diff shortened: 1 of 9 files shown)


controllers/reconcilers/grafana/deployment_reconciler.go
```diff
@@ -211,6 +220,41 @@ func getContainers(cr *v1beta1.Grafana, scheme *runtime.Scheme, vars *v1beta1.Op
...
+		Drop: []v1.Capability{
+			"ALL",
+		},
+	}
+	return &v1.SecurityContext{
+		AllowPrivilegeEscalation: model.BoolPtr(false),
+		ReadOnlyRootFilesystem:   model.BoolPtr(true),
+		Privileged:               model.BoolPtr(false),
+		RunAsNonRoot:             model.BoolPtr(true),
+		RunAsUser:                model.IntPtr(10001),
```
NissesSenap, line 235: We need to check if we are running on OCP before setting these default settings.

controllers/reconcilers/grafana/deployment_reconciler.go
```diff
@@ -211,6 +220,41 @@ func getContainers(cr *v1beta1.Grafana, scheme *runtime.Scheme, vars *v1beta1.Op
...
+	}
+}
+
+func getReadinessProbe(cr *v1beta1.Grafana) *v1.Probe {
+	return &v1.Probe{
+		ProbeHandler: v1.ProbeHandler{
+			HTTPGet: &v1.HTTPGetAction{
+				Path:   GrafanaHealthEndpoint,
+				Port:   intstr.FromInt(GetGrafanaPort(cr)),
+				Scheme: v1.URISchemeHTTP,
```
NissesSenap, line 247: Since we provide the entire readinessProbe object to the user they can easily overwrite the default http setting. But in general readinessProbe will use `http` and not `https`.

NissesSenap:
Need to verify what happens with these settings when adding plugins. I think we need to mount a disk which is an emptyDir on a special location so the plugins can be unpacked.

I will try to verify as well.

NissesSenap:
seccomp profle is missing, which should be set.

controllers/reconcilers/grafana/deployment_reconciler.go
HubertStefanski: Hmm, there's that or we can set validation to false, which would ignore unknown fields on the kubernetes custom resource so we wouldn't need to do env-specific logic, but that's also not an ideal situation, it can lead to misconfigurations quite easily. (so doing logic for this check is probably a must anwyay)

HubertStefanski (approved):

controllers/reconcilers/grafana/deployment_reconciler.go
weisdd: In v4, we do not allow users to specify scheme, it's chosen automatically based on the server.protocol setting. There, with the old approach of being able to specify all fields, we also had a CRD-based enforcement for the list of server protocols (only http and https).
I guess there are not many people who would run grafana with HTTPs backend, I would rather expect a transparent case of service mesh.

Not saying we need to change anything here, just adding the comment for the context.

controllers/reconcilers/grafana/deployment_reconciler.go
weisdd: How does it work in OCP nowadays? Does it mutate the `RunAsUser` field of individual pods? And when doing so, does it simply ignore the existing value?

And does it really need to be set? I mean wouldn't OCP consider an image with a `USER <non-root>` instruction set sufficient? - OCP doesn't work on m1, so can't check myself :)

weisdd:
One of the tests that we can do is to try deploying it a namespace with pod security standards enforced in `Restricted` mode ([docs](https://kubernetes.io/docs/concepts/security/pod-security-standards/)), that could help to quickly identify if there's any best-practice setting still missing.

controllers/reconcilers/grafana/deployment_reconciler.go
NissesSenap: @HubertStefanski yeah I think we want to keep validation, I reached out in the operator-sdk-dev channel in k8s and asked for the easiest workaround and got a reply: https://kubernetes.slack.com/archives/C017UU45SHL/p1671050125336449
OpenTelemetry collector operator seems to be doing something similar: https://github.com/open-telemetry/opentelemetry-operator/blob/b7081575f8742cddbf82b7e59c33b405882ba8af/pkg/autodetect/main.go#L74

So sadly OCP don't mutate the `RunAsUser` filed (unless this have changed the last 2 years).
Instead it just say that you can set RunAsUser due to SCC or something like that.

It needs to be set if we want to set `RunAsNonRoot` or some of the other you can't run as root rule. I don't remember which one it is.

controllers/reconcilers/grafana/deployment_reconciler.go
NissesSenap: Agreed, I think the limited amount of people using service mesh don't justify the reason to add any logic for this.
Thanks for the extra context.

controllers/reconcilers/grafana/deployment_reconciler.go
weisdd: @NissesSenap I think `RunAsNonRoot` only makes a pod (or a container?) fail in case container tries to start with `uid` 0, it does not specifically demand for `RunAsUser`. With it enabled, either of those two cases would satisfy the restriction:
- If an image has a `USER` instruction with a non-zero `uid` (or refers to a non-`root` username);
- If a container has `RunAsUser` (in its own spec or derived from pod's `securityContext`) with a non-zero `uid`.

```
~  docker run --rm -it --entrypoint sh grafana/grafana -c id
uid=472(grafana) gid=0(root) groups=0(root)
```

So, a custom `uid` is already chosen. From what I remember, the issue with OCP was that it required to use a specific range for uid which was different for each of the namespaces we had.

NissesSenap: Yes exactly so that is why we can't define any RunAsUser on OCP. I will write some logic to check for OCP and not set it.

NissesSenap:
So I broken out the platform detection functionality to a separate package just like they did in [opentelemetry-operator](https://github.com/open-telemetry/opentelemetry-operator/blob/b7081575f8742cddbf82b7e59c33b405882ba8af/pkg/autodetect/main.go#L74) and took more or less there whole solution.

This to be able to write a tests easier in the deployment reconciler (haven't written any so far). 

The good thing is that we don't have the same code at ingress and deployment, and we can use it any other controller that might need it.

The opentelemetry-operator uses a controller that constantly checks which version of HPA and if they are on OCP. I think that is a bit overkill and in general I think the operator would be restarted during a patching so I don't really see the need to check constantly through a controller. And currently we don't support HPA.

I took a way a bit of the extendibility that the opentelemetry-operator team have to instead regain some readability, I don't know if any other Kubernetes provider have a created a it's own ingress solution like OCP have. We at least haven't gotten any requests about it in version 4.
If it would happen we should be able to extend this functionality again without to much trouble.

In short how do you think we should proceed?
Looking forward to your comments.

NissesSenap:
So I think this PR is ready for a review.

I added seccompProfile defaults for the pod as well. 
I'm unable to verify in OCP so @HubertStefanski  or @pb82 if you can take it for a spin that would be great.

controllers/reconcilers/grafana/deployment_reconciler.go
```diff
@@ -211,7 +225,59 @@ func getContainers(cr *v1beta1.Grafana, scheme *runtime.Scheme, vars *v1beta1.Op
...
+		PeriodSeconds:       ReadinessProbePeriodSeconds,
+		SuccessThreshold:    ReadinessProbeSuccessThreshold,
+		FailureThreshold:    ReadinessProbeFailureThreshold,
+	}
+}
+
+func getPodSecurityContext() *v1.PodSecurityContext {
+	return &v1.PodSecurityContext{
+		SeccompProfile: &v1.SeccompProfile{
+			Type: "RuntimeDefault",
```
weisdd, line 275: I haven't had any experience running pods with SeccompProfile enabled, so my concerns are based on reading the docs:
- The functionality was moved to beta only 1.25, which is a recent release, so most users won't have it enabled by default;
- (from docs) "To use seccomp profile defaulting, you must run the kubelet with the SeccompDefault [feature gate](https://v1-25.docs.kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/) enabled (this is the default). You must also explicitly enable the defaulting behavior for each node where you want to use this with the corresponding --seccomp-default [command line flag](https://v1-25.docs.kubernetes.io/docs/reference/command-line-tools-reference/kubelet)."
  - Seems like a user must explicitly configure some things before it all starts kicking in.

So, I'd rather not add this field by default.

controllers/reconcilers/grafana/deployment_reconciler.go
```diff
@@ -13,38 +13,50 @@ import (
...
-	InitMemoryLimit   = "512Mi"
-	InitCpuLimit      = "1000m"
-	MemoryRequest     = "256Mi"
-	CpuRequest        = "100m"
-	MemoryLimit       = "1024Mi"
-	CpuLimit          = "500m"
+	InitMemoryRequest                       = "128Mi"
+	InitCpuRequest                          = "250m"
+	InitMemoryLimit                         = "512Mi"
+	InitCpuLimit                            = "1000m"
```
weisdd, line 26: Do we actually need so much cpu power for the init container?

controllers/reconcilers/grafana/deployment_reconciler.go
weisdd: Also:
- "It is possible that the default profiles differ between container runtimes and their release versions, for example when comparing those from CRI-O and containerd."

NissesSenap: So I had to read the docs again and the possibility to use seccompprofile was made GA in 1.19.
https://kubernetes.io/docs/tutorials/security/seccomp/

The new feature that becomes beta in 1.25 is that we don't have to set this value, instead it will apply by default to all pods unless you specifically say no.
When this feature becomes GA we can remove this config from the operator, sadly we have to wait for GA since the enablement of beta features was disabled in 1.25 or 1.26 (can't remember which one).

Containerd, docker and cri-o all have defaultprfoiles and I assume all CNI:s do.
And we can see in the docs of OCP that they even apply the default seccompprofile after OCP 4.11 https://docs.openshift.com/container-platform/4.11/security/seccomp-profiles.html
But they use the old annotation solution, but they can be used at the same time, at least from a kubernetes point of view.

So I don't see any issues with adding seccompprofile as default the grafana instance.

controllers/reconcilers/grafana/deployment_reconciler.go
NissesSenap: Agreed, we can do a separate PR to fix that.

controllers/reconcilers/grafana/deployment_reconciler.go
weisdd: @NissesSenap Ah, I guess I misunderstood the docs then, sorry for that :)
If that's the case, then I guess we can add it back :)

NissesSenap: No worries at all :)

examples/grafana_deployment/resources.yaml
```diff
@@ -0,0 +1,29 @@
+---
+apiVersion: grafana.integreatly.org/v1beta1
+kind: Grafana
+metadata:
+  name: grafana
+  labels:
+    dashboards: "grafana"
+spec:
+  client:
+    preferIngress: true
```
weisdd, line 10: We need to remove references to ingress :)
```suggestion
```

main.go
```diff
@@ -97,10 +98,22 @@ func main() {
...
 
+	restConfig := ctrl.GetConfigOrDie()
+	autodetect, err := autodetect.New(restConfig)
+	if err != nil {
+		setupLog.Error(err, "failed to setup auto-detect routine")
+		os.Exit(1)
+	}
+	isOpenShift, err := autodetect.IsOpenshift()
+	if err != nil {
+		setupLog.Error(err, "unable to check platform")
```
weisdd, line 109: Should we also terminate the controller if it's unable to detect the platform?

weisdd (changes requested):

PR #51, opened 2022-12-18
Kuttl e2e solution

NissesSenap:
I gave this a try in version 4 but we didn't have a great usage of status fields so we had to run a bunch of ugly scripts inside kuttl to make it work. This time we will be able to verify things through status fields.


NissesSenap:
## TODO

- setup a github actions pipeline that uses the new e2e tests.
- verfiy that installation of kuttle using the makefile etc works as it should
- test grafanafolder
- test some more uses-cases like multiple grafana instances.

NissesSenap:
The PR finally passes, don't know why it won't trigger in this PR but it works in my own PR: https://github.com/NissesSenap/grafana-operator-experimental/pull/2

So this should be ready for review.

HubertStefanski:
> The PR finally passes, don't know why it won't trigger in this PR but it works in my own PR: [NissesSenap#2](https://github.com/NissesSenap/grafana-operator-experimental/pull/2)
> 
> So this should be ready for review.

Could this be due to the fact that it's running against a fork? I'm not sure how GH actions handle this. However, since you're a maintainer on the repository it should run fine :thinking:

config/manager/kustomization.yaml
```diff
@@ -12,5 +12,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: controller
-  newTag: latest
+  newName: quay.io/grafana-operator/grafana-operator
+  newTag: v5.0.0
```
HubertStefanski, line 16: Can we jinja template ${VERSION} here instead? :thinking:

HubertStefanski (approved):

config/manager/kustomization.yaml
NissesSenap: So this is actually solved through the makefile
`IMG ?= quay.io/grafana-operator/grafana-operator:$(VERSION)`
So when `make deploy` it will generate the file for us so we don't have to think of it :).

HubertStefanski: Ahhh, sweet! :+1:

PR #52, opened 2022-12-18
Grafana rbac, deployment ingress etc.

NissesSenap:


config/default/kustomization.yaml
```diff
@@ -1,5 +1,5 @@
 # Adds namespace to all resources.
-namespace: grafana-operator-experimental-system
```
NissesSenap, line 2: Since we will merge in to the grafana-operator it will just create issues if we call it experimental when we are developing. Especially if we will write e2e tests etc.

NissesSenap:
I don't know how we should manage cluster vs namespace mode from a kubebuilder point of view.
I will merge this PR now since I need it for other things and I think this is a good start.

In general we need to start playing around with the operator without running it through the Makefile.
I started writing e2e tests and that was when I found this issue.

PR #56, opened 2022-12-19
fix: reconcile only when a stage is not finished

pb82:
Reconciler only when a stage either failed or is not finished yet, while still making sure that the status is updated with an error message if needed:

1. If an error occurs or a reconciler returns `in progress`, the grafana controller will retry in 10 seconds.
2. The status is updated only if there are actual changes (e.g. an error message). This was already the case.
3. If reconcile was successful and if there was no change in status, the status is not updated.


HubertStefanski (approved):
LGTM!

PR #57, opened 2022-12-21
feat: allow importing dashboards that define inputs and map them to a datasource

pb82:
adds a new field to the dashboard spec that allows mapping internal data sources to inputs defined on the dashboard.


examples/datasource_mapping/resources.yaml
```diff
@@ -0,0 +1,241 @@
...
+    type: prometheus
+    access: proxy
+    url: http://prometheus-service:9090
+    isDefault: true
+    jsonData:
+      'tlsSkipVerify': true
+      'timeInterval': "5s"
+    editable: true
+---
+apiVersion: grafana.integreatly.org/v1beta1
```
NissesSenap, line 39: Why do we need the first sampledatasource?
Currently it only confuses me.

api/v1beta1/grafanadashboard_types.go
```diff
@@ -32,9 +32,14 @@ type DashboardSourceType string
 const (
 	DashboardSourceTypeRawJson DashboardSourceType = "json"
 	DashboardSourceTypeUrl     DashboardSourceType = "url"
-	DefaultResyncPeriod                            = "24h"
```
NissesSenap, line 35: note the change of the defaultResyncPeriod.

NissesSenap (commented):
LGTM, added a minor comment about the example.
Maybe not a bad thing to add some extra info in the README.

examples/datasource_mapping/resources.yaml
pb82: The intention was to demonstrate that you can select a non-default datasource with this mapping. If there was only one data source, I believe it would be picked by default.

examples/datasource_mapping/resources.yaml
```diff
@@ -0,0 +1,241 @@
...
+kind: Grafana
+metadata:
+  name: grafana
+  labels:
+    dashboards: "grafana"
+spec:
+  client:
+    preferIngress: true
```
weisdd, line 10: The example is not entirely correct given the fact that now operator expects ingress spec. I think, in general case, it's easier not to set `preferIngress` at all.

pb82: ah, good point. Will remove this.

examples/datasource_mapping/resources.yaml
NissesSenap: I wrote a small note about it in the README.

examples/datasource_mapping/resources.yaml
NissesSenap: I fixed this in the example.

PR #60, opened 2022-12-28
Remove PreserveUnknownFields from CRD

NissesSenap:
I don't see why we should use `PreserveUnknownFields` it will only confuse people when they apply something but something potentially not happening.

It's better that the config is removed when they check the CR.


pb82:
preserving unknown fields is required in those cases

PR #61, opened 2022-12-30
Datasource env vars

pb82:
Adds a secrets list to the datasource spec and allows expanding variables in the datasource yaml.


controllers/reconcilers/grafana/ingress_reconciler.go
```diff
@@ -123,6 +123,22 @@ func (r *IngressReconciler) getIngressAdminURL(ingress *v1.Ingress) string {
 		}
 	}
 
+	// if all fails, try to get access trough the load balancer
+	var loadBalancerIp = ""
+	for _, lb := range ingress.Status.LoadBalancer.Ingress {
+		if lb.Hostname != "" {
+			hostname = lb.Hostname
```
weisdd, line 130: I think this part should be rewritten. Say, hostname is defined in one of `ingress.Spec.Rules` and you also have a hostname in `ingress.Status.LoadBalancer.Ingress`. With the current implementation, the latter would take precedence whereas we have to rather rely on the former.
As an option, you could reuse the same variable name for either of the two (`lb.Hostname`, `lb.IP`) and then save the value into `hostname` if `hostname` is still empty by that point.

weisdd (changes requested):
Thanks for implementing it! I thought it would require a more sophisticated logic, happy to see it being something straightforward.

controllers/datasource_controller.go
```diff
@@ -280,9 +285,14 @@ func (r *GrafanaDatasourceReconciler) onDatasourceCreated(ctx context.Context, g
 		return err
 	}
 
+	variables, err := r.CollectVariablesFromSecrets(ctx, cr)
+	if err != nil {
```
NissesSenap, line 289: Could it be wroth writing some custom error or even update the status of the datasource that the controller was unable to get the secret or something like that?

NissesSenap (commented):
Great job. Minor comment from me.

controllers/datasource_controller.go
pb82: Currently you would see the error message only in the operator logs. But I agree that adding it to the status would be a nicer UX.

api/v1beta1/grafanadatasource_types.go
```diff
@@ -146,6 +156,26 @@ func (in *GrafanaDatasource) Unchanged() bool {
 	return in.Hash() == in.Status.Hash
 }
 
+func (in *GrafanaDatasource) ExpandVariables(variables map[string][]byte) ([]byte, error) {
+	if in.Spec.Datasource == nil {
```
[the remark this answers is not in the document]
pb82: :+1:

controllers/reconcilers/grafana/ingress_reconciler.go
pb82: that's a good point. I was adding this code because otherwise I couldn't test my change in k3d. I'll make sure that it's only applied if the hostname couldn't be determined before.

pb82:
@NissesSenap @weisdd all comments are now addressed

NissesSenap (approved):

PR #62, opened 2022-12-30
go linting and basic pr validation

NissesSenap:
Started to write some CI stuff to force us to get to a decent level before we create our first rc.

A few fixes around [implicit memory aliasing in for loop](https://stackoverflow.com/questions/62446118/implicit-memory-aliasing-in-for-loop).

Wrote a really ugly test around it in the dashboard controller to make it visible.
To tiered to make it better right now.
There is probably a simple way of doing it, if you anyone sees a simple way of doing it please share :)

### DIFF
--- a/.github/workflows/pr-validation.yaml
+++ b/.github/workflows/pr-validation.yaml
@@ -30,3 +30,24 @@ jobs:
         uses: golangci/golangci-lint-action@v3.4.0
         with:
           version: "v1.51.2"
+
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone repo
+        uses: actions/checkout@755da8c3cf115ac066823e79a1e1788f8940201b #v3.2.0
+      - name: Setup go
+        uses: actions/setup-go@v3
+        with:
+          go-version-file: "go.mod"
+      - name: Run test
+        shell: bash
+        run: |
+          make test
+      - name: Check if working tree is dirty
+        run: |
+          if [[ $(git status --porcelain) ]]; then
+            git diff
+            echo 'run make test and commit changes'
+            exit 1
+          fi
--- a/.github/workflows/pr-validation.yaml
+++ b/.github/workflows/pr-validation.yaml
@@ -6,14 +6,15 @@ name: PR Validation
 on: pull_request
 
 jobs:
-  tests:
-    name: file-checks
+  file-checks:
     runs-on: ubuntu-latest
     steps:
       - name: Clone the code
         uses: actions/checkout@755da8c3cf115ac066823e79a1e1788f8940201b #v3.2.0
+
       - name: trailing space
         uses: chainguard-dev/actions/trailing-space@main
+
       - name: eof-check
         uses: chainguard-dev/actions/eof-newline@main
 
@@ -22,10 +23,14 @@ jobs:
     steps:
       - name: Clone repo
         uses: actions/checkout@755da8c3cf115ac066823e79a1e1788f8940201b #v3.2.0
+        with:
+          submodules: recursive
+
       - name: Setup go
         uses: actions/setup-go@v3
         with:
           go-version-file: "go.mod"
+
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v3.4.0
         with:
@@ -36,14 +41,18 @@ jobs:
     steps:
       - name: Clone repo
         uses: actions/checkout@755da8c3cf115ac066823e79a1e1788f8940201b #v3.2.0
+        with:
+          submodules: recursive
+
       - name: Setup go
         uses: actions/setup-go@v3
         with:
           go-version-file: "go.mod"
+
       - name: Run test
-        shell: bash
         run: |
           make test
+
       - name: Check if working tree is dirty
         run: |
           if [[ $(git status --porcelain) ]]; then
--- a/.github/workflows/pr-validation.yaml
+++ b/.github/workflows/pr-validation.yaml
@@ -5,9 +5,6 @@ name: PR Validation
 
 on: pull_request
 
-env:
-  GO_VERSION: 1.19
-
 jobs:
   docs_only_check:
     name: Check for docs-only change
@@ -19,8 +16,6 @@ jobs:
     steps:
     - name: Check out code
       uses: actions/checkout@755da8c3cf115ac066823e79a1e1788f8940201b #v3.2.0
-      with:
-        fetch-depth: 2
     - id: files
       name: Get changed files
       uses: tj-actions/changed-files@0626c3f94002c0a9d7491dd7fed7055bbdff6f92 #v35.1.0
@@ -58,7 +53,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v3
         with:
-          go-version: ${{ env.GO_VERSION }}
+          go-version-file: "go.mod"
           check-latest: true
       - name: go fmt
         uses: chainguard-dev/actions/gofmt@main
@@ -73,8 +68,8 @@ jobs:
       - name: Setup go
         uses: actions/setup-go@v3
         with:
-          go-version: ${{ env.GO_VERSION }}
+          go-version-file: "go.mod"
       - name: golangci-lint
-        uses: golangci/golangci-lint-action@v3.3.1
+        uses: golangci/golangci-lint-action@v3.4.0
         with:
-          version: "v1.50.1"
+          version: "v1.50.2"
--- a/.github/workflows/pr-validation.yaml
+++ b/.github/workflows/pr-validation.yaml
@@ -6,59 +6,16 @@ name: PR Validation
 on: pull_request
 
 jobs:
-  docs_only_check:
-    name: Check for docs-only change
-    runs-on: ubuntu-latest
-    permissions:
-      contents: read
-    outputs:
-      docs_only: ${{ steps.docs_only_check.outputs.docs_only }}
-    steps:
-    - name: Check out code
-      uses: actions/checkout@755da8c3cf115ac066823e79a1e1788f8940201b #v3.2.0
-    - id: files
-      name: Get changed files
-      uses: tj-actions/changed-files@0626c3f94002c0a9d7491dd7fed7055bbdff6f92 #v35.1.0
-      with:
-        files_ignore: '**.md'
-    - id: docs_only_check
-      if: steps.files.outputs.any_changed != 'true'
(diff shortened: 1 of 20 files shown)


controllers/model/grafana_resources.go
```diff
@@ -20,7 +20,7 @@ func GetGrafanaConfigMap(cr *grafanav1beta1.Grafana, scheme *runtime.Scheme) *v1
 			Namespace: cr.Namespace,
 		},
 	}
-	controllerutil.SetOwnerReference(cr, config, scheme)
+	controllerutil.SetOwnerReference(cr, config, scheme) //nolint:errcheck
```
NissesSenap, line 23: Not sure if we should ignore these errors everywhere, it currently works but it would also be very strange if we can't set ownership (as long as nothing else is owning the resource it shoulden't be a problem.
But you can actually remove for example a pods owner but I assume that the controller re-adds it directly, so we still need the functionality of setting it during a normal reconciliation.

At the same time the ownership should be set during its initial creation.
Maybe it's better to merge this first and then potentially do a refactoring of that

.golangci.yaml
```diff
@@ -0,0 +1,135 @@
...
+    - nakedret
+    - prealloc
+    - nolintlint
+    - staticcheck
+    - thelper
+    - whitespace
+
+linters-settings:
+  gocyclo:
+    min-complexity: 25
```
NissesSenap, line 34: Some of these settings is rather high, but I want to get this PR merged. We can work on lovering these setting at a later stage.

NissesSenap (commented):
Far from the perfect PR but at least a start with linting

.golangci.yaml
```diff
@@ -0,0 +1,135 @@
...
+    - nilerr
+    - unparam
+    - gosec
+    - unused
+    - govet
+    - gosimple
+    - errorlint
+    - errcheck
+    - dogsled
+    - cyclop
```
weisdd, line 18: There's a certain intersection between what `gocyclo` and `cyclop` offer. The latter seems to no longer be maintained, so maybe we need only `gocyclo`.

.golangci.yaml
```diff
@@ -0,0 +1,135 @@
...
+    - exhaustive
+    - nestif
+    - goconst
+    - gofmt
```
weisdd, line 22: I think it's worth switching to `gofumpt` here as it contains more formatting rules (backwards compatible with `gofmt`) and import sorting. I've been using it for quite a while, very happy with the results.

.golangci.yaml
```diff
@@ -0,0 +1,135 @@
...
+  disable-all: false
+  enable:
+    - gocyclo
+    - misspell
+    - nilerr
+    - unparam
+    - gosec
+    - unused
+    - govet
+    - gosimple
```
weisdd, line 14: I would also add `gocritic`.

.github/workflows/pr-validation.yaml
```diff
@@ -0,0 +1,80 @@
...
+      - name: Setup Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: ${{ env.GO_VERSION }}
+          check-latest: true
+      - name: go fmt
+        uses: chainguard-dev/actions/gofmt@main
+        with:
+          args: "-s"
+
```
weisdd, line 67:
```suggestion
```

This is redundant. If code is not properly formatted, then golangci-lint would complain assuming either `gofmt` or `gofumpt` is enabled in its config.

.github/workflows/pr-validation.yaml
```diff
@@ -0,0 +1,80 @@
...
+    - id: files
+      name: Get changed files
+      uses: tj-actions/changed-files@0626c3f94002c0a9d7491dd7fed7055bbdff6f92 #v35.1.0
+      with:
+        files_ignore: '**.md'
+    - id: docs_only_check
+      if: steps.files.outputs.any_changed != 'true'
+      name: Check for docs-only changes
+      run: echo "docs_only=true" >> $GITHUB_OUTPUT
+
```
weisdd, line 33:
```suggestion
```

Since `fmt` job is redundant, `docs_only_check` is not needed either.

.github/workflows/pr-validation.yaml
```diff
@@ -0,0 +1,80 @@
...
+  go-lint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone repo
+        uses: actions/checkout@755da8c3cf115ac066823e79a1e1788f8940201b #v3.2.0
+      - name: Setup go
+        uses: actions/setup-go@v3
+        with:
+          go-version: ${{ env.GO_VERSION }}
```
weisdd, line 76:
```suggestion
          go-version-file: 'go.mod'
```
I'd rather let it be in sync with whatever is mentioned in `go.mod`.

.github/workflows/pr-validation.yaml
```diff
@@ -0,0 +1,80 @@
...
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v3.3.1
+        with:
+          version: "v1.50.1"
```
weisdd, line 80:
```suggestion
        uses: golangci/golangci-lint-action@v3.4.0
        with:
          version: "v1.50.2"
```
I think we can bump the tags right away.

.github/workflows/pr-validation.yaml
```diff
@@ -0,0 +1,80 @@
...
+      run: echo "docs_only=true" >> $GITHUB_OUTPUT
+
+  tests:
+    name: file-checks
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone the code
+        uses: actions/checkout@755da8c3cf115ac066823e79a1e1788f8940201b #v3.2.0
+        with:
+          fetch-depth: 0
```
weisdd, line 41:
```suggestion
        uses: actions/checkout@3
```

I think there shouldn't be much need in pinning this action by the commit.
Also, there's no need to run it with custom fetch-depth.

.github/workflows/pr-validation.yaml
```diff
@@ -0,0 +1,80 @@
+permissions:
+  contents: read
+
+name: PR Validation
+
+on: pull_request
+
+env:
+  GO_VERSION: 1.19
+
```
weisdd, line 10:
```suggestion
```
Not needed if suggestion with `go-version-file` is accepted.

weisdd (changes requested):

.github/workflows/pr-validation.yaml
NissesSenap: Didn't know that it existed, really nice.

.github/workflows/pr-validation.yaml
NissesSenap: So the fetch thing was a copy-paste from the current grafana-operator repo where we needed it to be able to get the submodule for jsonet but it shouldn't  be needed in the new version so I remove it.

The reason why I use the commit sha is due to that I have thoughts about making our repo salsa compliant https://slsa.dev/ and I have copied a number of things from chainguard that knows their stuff around this. But in short it's due to tags can be rewritten but commits can't so we can trust a commit be consistent.

But I have removed the fetch-depth.

NissesSenap: We can merge this one first, and in @HubertStefanski PR we can fix it.

controllers/model/dashboard_resources.go
```diff
@@ -17,17 +17,6 @@ func GetPluginsConfigMap(cr *grafanav1beta1.Grafana, scheme *runtime.Scheme) *v1
 			Namespace: cr.Namespace,
 		},
 	}
-	controllerutil.SetOwnerReference(cr, config, scheme)
-	return config
-}
-
-func GetDashboardsConfigMap(cr *grafanav1beta1.Grafana, scheme *runtime.Scheme) *v1.ConfigMap {
```
NissesSenap, line 24: This feature isn't used I'm unsure around the logic here, is dashboards supposed to be in a configmap @pb82 ?

Makefile
```diff
@@ -278,3 +278,22 @@ KUTTL=$(GOBIN)/kuttl
 else
 KUTTL=$(shell which kubectl-kuttl)
 endif
+
+golangci:
+ifeq (, $(shell which golangci-lint))
+	@{ \
+	set -e ;\
+	KUTTL_TMP_DIR=$$(mktemp -d) ;\
```
weisdd, line 286: I think we should not refer to kuttl here. Plus, do we actually need to do go mod init? - go install should work without that just fine.

controllers/dashboard_controller_test.go
```diff
@@ -0,0 +1,121 @@
...
+
+import (
+	"testing"
+
+	"github.com/grafana-operator/grafana-operator-experimental/api/v1beta1"
+	"github.com/stretchr/testify/assert"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var (
```
weisdd, line 27: I think it'd be better to either move these vars into the test function itself or wrap them into a helper function, otherwise the vars will leak into other tests.

controllers/dashboard_controller_test.go
```diff
@@ -0,0 +1,121 @@
...
+					"grafana-operator-system/broken2/cb1688d2-547a-465b-bc49-df3ccf3da883",
+				},
+			},
+		},
+	}
+)
+
+func TestGetDashboardsToDelete(t *testing.T) {
+	dashboardsToDelete := getDashboardsToDelete(&dashboardList, grafanaList)
+	for tt := range dashboardsToDelete {
```
weisdd, line 113: I think it'd be better to use another name instead of tt. tt normally implies a test case, whereas here it's just an element of a map.

controllers/dashboard_controller_test.go
```diff
@@ -0,0 +1,121 @@
...
+		if tt.Name == "internal-broken1" {
+			assert.Equal(t, dashboardsToDelete[tt], []v1beta1.NamespacedResource([]v1beta1.NamespacedResource{"grafana-operator-system/broken1/cb1688d2-547a-465b-bc49-df3ccf3da883"}))
```
weisdd, line 115: Here and in another assertion, we need to swap the values, because expected value has to come first.
(`func assert.Equal(t assert.TestingT, expected interface{}, actual interface{}, msgAndArgs ...interface{}) bool`)

weisdd (changes requested):

weisdd (approved):
LGTM

PR #70, opened 2023-01-18
WIP: jsonnet support

HubertStefanski:
Addresses !67


controllers/fetchers/jsonnet_fetcher.go
```diff
@@ -0,0 +1,15 @@
...
+import (
+	"github.com/google/go-jsonnet"
+	"github.com/grafana-operator/grafana-operator-experimental/api/v1beta1"
+)
+
+func FetchJsonnet(dashboard *v1beta1.GrafanaDashboard) (string, error) {
+
+	vm := jsonnet.MakeVM()
+
+	vm.Importer(&jsonnet.FileImporter{JPaths: []string{"/opt/jsonnet"}})
```
pb82, line 12: is this path meant for jsonnet libraries?

controllers/fetchers/jsonnet_fetcher.go
```diff
@@ -0,0 +1,15 @@
+package fetchers
+
+import (
+	"github.com/google/go-jsonnet"
+	"github.com/grafana-operator/grafana-operator-experimental/api/v1beta1"
+)
+
+func FetchJsonnet(dashboard *v1beta1.GrafanaDashboard) (string, error) {
+
+	vm := jsonnet.MakeVM()
```
pb82, line 10: I wonder if there's a way to have a global VM that we reset after every import?

HubertStefanski: Hmmm, that's not a bad idea.
Do you think it's a safe assumption to make that if one dashboard is using jsonnet, that the remainder will be as well?

controllers/reconcilers/grafana/config_reconciler.go
```diff
@@ -37,8 +40,31 @@ func (r *ConfigReconciler) Reconcile(ctx context.Context, cr *v1beta1.Grafana, s
 		configMap.Data["grafana.ini"] = config
 		return nil
 	})
+
+	if cr.Spec.Config["grafonnet"]["location"] != "" {
+		err = r.reconcileGrafonnetConfigMaps(*configMap)
+		if err != nil {
+			return v1beta1.OperatorStageResultFailed, err
+		}
+	}
```
HubertStefanski, line 49: Needs to be dropped

controllers/reconcilers/grafana/config_reconciler.go
```diff
@@ -37,8 +40,31 @@ func (r *ConfigReconciler) Reconcile(ctx context.Context, cr *v1beta1.Grafana, s
...
+
+	for filename, contents := range cfg.Data {
+		filePath := fmt.Sprintf("%v/%v", folderPath, filename)
+		err = utils.CreateFile(filePath, contents)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
```
HubertStefanski, line 70: Unused, needs to be dropped

utils/grafonnet_utils.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+func validateFileExtension(filePath string) error {
+	//check for a valid jsonnet extension
+	extension := filepath.Ext(filePath)
+	if extension != JsonnetExtension {
+		return fmt.Errorf("unknown extension, expected %v", JsonnetExtension)
+	}
+	return nil
+}
```
HubertStefanski, line 47: Will either be dropped or used in the jsonnet fetcher, i'm still deciding what a cleaner way is

HubertStefanski:
Seeing an unexpected EOF error when running locally, still troubleshooting.
This has to do with the loaded jsonnet from the go embed call. 
Still unclear why it's not writing with a proper EOF, `\n` at the end of the embed/grafana.libsonnet  file has no effect

controllers/fetchers/jsonnet_fetcher.go
pb82: Not sure, thinking about it again, maybe its not expensive to recreate a fresh VM and this way we also provide isolation between dependencies used by dashboards.

HubertStefanski:
Mostly done and working, still need to find out how this will work on a cluster deployment, It might need a copy in the containerfile. I'll do that tomorrow

controllers/fetchers/jsonnet_fetcher.go
HubertStefanski: Well, I guess we can refactor to a more permanent VM if there's a big use case for jsonnet, but for now, wasteful to provision it for a single dashboard type if there's a mixture of dashboard types.

Albeit, the person using jsonnet, would probably use jsonnet for all their dashboards.

HubertStefanski:
@pb82 @weisdd @NissesSenap I think this is ready to merge, mind taking a look?

pb82:
Thanks @HubertStefanski this is great! I'll try it out.

controllers/fetchers/jsonnet_fetcher.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+func FetchJsonnet(dashboard *v1beta1.GrafanaDashboard, libsonnet embed.FS, libsonnetPath string) ([]byte, error) {
+	vm := jsonnet.MakeVM()
+
+	if dashboard.Spec.Jsonnet == "" {
+		return nil, errors.New("no jsonnet Content Found, nil or empty string")
+	}
+
+	_, err := os.Stat(libsonnetPath)
```
weisdd, line 21: Like you mentioned, this will need to be adjusted as `os` package looks into "real" filesystem, not into the virtualized one.
As I see, `embed` implements Stat() ([link](https://cs.opensource.google/go/go/+/refs/tags/go1.20:src/embed/embed.go;drc=15605ca827723d568f780402d03b29842fd20eec;l=353)), so I guess you need to replace it with fs.Stat or something like that :)

controllers/config/operator_constants.go
```diff
@@ -11,6 +11,9 @@ const (
 	GrafanaPluginsPath      = "/var/lib/grafana/plugins"
 	GrafanaProvisioningPath = "/etc/grafana/provisioning/"
 
+	// Grafonnet
+	GrafonnetEmbedPath = "./embeds/grafonnet/grafonnet-lib/"
```
weisdd, line 15: Would be nice to add a note somewhere that would explain where to get the library from, which files to copy, and so on. - Just to make our life easier in an event when we need to update the library.

weisdd (changes requested):

controllers/config/operator_constants.go
HubertStefanski: I added readme in the embeds dir. I also switched to using a gitmodule, which means updating the lib is just `git submodule update` no longer any need to manually move things around. 

ps. I've edited the .gitignore to hopefully prevent overbloating the grafonnet-lib directory

Dockerfile
```diff
@@ -27,7 +28,9 @@ RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o ma
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
 FROM gcr.io/distroless/static:nonroot
 WORKDIR /
+
 COPY --from=builder /workspace/manager .
+COPY embeds/ embeds/
```
weisdd, line 33: I syspect this should not be required since you're embedding everything in the binary.

embeds/grafana_embeds.go
```diff
@@ -0,0 +1,12 @@
+package embeds
+
+import "embed"
+
+//go:embed grafonnet-lib
+var GrafonnetEmbed embed.FS
+
+//go:embed testing/dashboard.jsonnet
+var TestDashboardEmbed []byte
```
pb82, line 9: Is this needed? We already have an examples directory.

embeds/grafana_embeds.go
```diff
@@ -0,0 +1,12 @@
...
+
+//go:embed testing/dashboard.json
+var TestDashboardEmbedExpectedJSON []byte
```
pb82, line 12: See above.

pb82:
@HubertStefanski can we include the git submodule for Grafonnet in the repo? Or at least instructions for how to fetch Grafonnet?

Dockerfile
pb82: +1

pb82 (approved):
I've verified that importing jsonnet/grafonnet dashboards works :+1:

.github/workflows/e2e.yaml
```diff
@@ -15,7 +15,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Clone repo and checkout submodules
-        uses: actions/checkout@v3.3.0
+        uses: actions/checkout@v3.3.0:
+      - name: Checkout submodules
+        run: git submodule update --init --recursive
```
weisdd, line 20:
```suggestion
        uses: actions/checkout@v3.3.0:
        with:
          submodules: recursive
```
That would be easier :)

PR #72, opened 2023-02-26
feat: deployment options using kustomize

pb82:
Deployment options using kustomize:

1. specify the target namespace in `deploy/base/kustomization.yaml`
2. For a namespace scoped installation, run `kubectl apply -k deploy/overlays/namespace_scoped`
3. For a cluster scoped installation, run `kubectl apply -k deploy/overlays/cluster_scoped`

Note: this requires a recent `kubectl` with kustomize and replacements support.


Makefile
```diff
@@ -3,7 +3,10 @@
 ###
 
 # Image URL to use all building/pushing image targets
-IMG ?= controller:latest
+VERSION ?= 5.0.0
+REGISTRY ?= quay.io
+REPO ?= pb82
```
NissesSenap, line 8: This was probably from your testing.

deploy/base/serviceaccount.yaml
```diff
@@ -0,0 +1,6 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: grafana-operator-controller-manager
+  namespace: default
```
NissesSenap, line 6: probably better to remove namespace or use a better default value

deploy/base/namespace.yaml
```diff
@@ -0,0 +1,5 @@
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: default
```
NissesSenap, line 5: Go with better default settings in base. Like grafana-operator-system or grafana

deploy/overlays/namespace_scoped/kustomization.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+replacements:
+  - source:
+      kind: Deployment
+      fieldPath: metadata.namespace
+    targets:
+      - select:
+          kind: Deployment
+        fieldPaths:
+          - spec.template.spec.containers.[name=manager].args.3
```
NissesSenap, line 15: Feels rather risky to write it like this since if we change anything in base it will the order of the args can get changed.
I would probably just patch the entire args fields to be sure.

NissesSenap (commented):
Nice work @pb82 , I added a few minor comments.

Makefile
pb82: Oops sorry, will reset to the official repo.

deploy/base/serviceaccount.yaml
pb82: @NissesSenap this is required by kustomize in order to replace the namespace. It will apply whichever namespace is set in `base/kustomization.yaml`, but for some resources (role, rolebinding) the namespace *must* be set to default, otherwise it is not replaced. So I thought to just use default everywhere.

deploy/overlays/namespace_scoped/kustomization.yaml
pb82: Yes, this is very brittle but I couldn't find a better way to set the value of a container arg in an overlay. Do you know if there is a better way? The goal is to:

- set `--namespace=<namespace>` in namespaced mode
- set omit `--namespace` or set to empty string in cluster scoped mode

deploy/overlays/namespace_scoped/kustomization.yaml
```diff
@@ -0,0 +1,13 @@
+bases:
+  - ../../base
+
+resources:
+  - rbac.yaml
+
+patches:
```
NissesSenap, line 7: This is the patch

deploy/overlays/namespace_scoped/deployment.yaml
```diff
@@ -0,0 +1,14 @@
...
+spec:
+  template:
+    spec:
+      containers:
+        - name: manager
+          args:
+            - --health-probe-bind-address=:8081
+            - --metrics-bind-address=127.0.0.1:8080
+            - --leader-elect
+            - --namespace=true
```
NissesSenap, line 14: @pb82 I pushed a commit with the patches usage. But I saw your comment from earlier.
I thought the usage was just `--namespace=true` but reading your comment it needs to be `--namespace=grafana-operator` or which ever namespace you want to monitor?

Is there a reason for this? Can't we assume that if the user wants to only deploy to one namespace we can also assume that he wants the grafana instance to be together with the operator.
Else you still need to write a clusterrole for the operator, sure you would still only do a rolebinding but still.

From the controller I assume it's rather easy to check witch namespace it's deployed to.

NissesSenap: If we want to keep this to a similar way we could use the downward API, but then we would have to change over to a env variable.

env:
    - name: namespace
      valueFrom:
        fieldRef:
          fieldPath: metadata.namespace

NissesSenap (commented):
I submitted a commit that uses the patch.

deploy/overlays/namespace_scoped/deployment.yaml
pb82: True, if we always assume that `--namespace` means the Operator namespace, we wouldn't have to worry about injecting the namespace into the args somehow (and could also just use a patch instead of a replacement).

config/manager/manager.yaml
```diff
@@ -53,5 +53,10 @@ spec:
           requests:
             cpu: 100m
             memory: 20Mi
+        env:
```
NissesSenap, line 56: This will always set the WATCH_NAMESPACES to the place where you deploy the operator.
Or am I thinking wrong?

deploy/base/deployment.yaml
```diff
@@ -0,0 +1,66 @@
...
+              name: https
+              protocol: TCP
+          resources: { }
+        - args:
+            - --health-probe-bind-address=:8081
+            - --metrics-bind-address=127.0.0.1:8080
+            - --leader-elect
+          command:
+            - /manager
+          image: quay.io/pb82/grafana-operator:5.0.0
```
NissesSenap, line 39: Still some pb82 thing remaining

NissesSenap (commented):
Minor comments.

config/manager/manager.yaml
pb82: It seems like it, but when I run `make bundle` with this, it actually sets the watch namespace to `metadata.annotations['olm.targetNamespaces']` in the CSV. But I'll change it here as well.

NissesSenap (approved):

PR #74, opened 2023-02-27
Initial helm chart

NissesSenap:
I think it would be nice if we can support a basic helm chart to deploy the operator.

This chart also pushes the container images to ghcr, I only see it as a good thing to support both quay and ghcr.
I have no issues with making quay the default one.


HubertStefanski:
Hmm, would we support this install method on top of what the community started here? https://github.com/bitnami/charts/tree/main/bitnami/grafana-operator

.github/workflows/release-github.yaml
```diff
@@ -0,0 +1,84 @@
...
+        run: |
+          echo "tag=${GITHUB_REF#refs/tags/}" >> $GITHUB_OUTPUT
+      - name: Cache container layers
+        uses: actions/cache@v2.1.4
+        with:
+          path: /tmp/.buildx-cache
+          key: ${{ runner.os }}-buildx-${{ github.sha }}
+          restore-keys: |
+            ${{ runner.os }}-buildx-
+      - name: Set up QEMU
```
weisdd, line 50: QEMU is slow as hell when trying to build an image for other architectures. When I tested it for a tiny app with not much dependencies, it took 4m28s instead of 51s. With the operator, it's likely to be much more.
The fastest option is to build an image for different architectures on the host and then copy binaries. I think `go-releaser` or `ko` would be much better in that sense. WDYT?

NissesSenap:
> Hmm, would we support this install method on top of what the community started here? https://github.com/bitnami/charts/tree/main/bitnami/grafana-operator

Yes that is my plan, I know we haven't spoken about this internally so if you don't think it's a good idea we can just scrap the idea, I haven't spent to much time on this.

But it feels more natural to do so, especially if we will start to sign artifcats.
And I have been thinking lots around it and I don't think the maintenance burdon will be to bad.

.github/workflows/release-github.yaml
NissesSenap: Saw now that my build of these images in my fork took 40 minutes to run....

weisdd: Yeah, that's what I was expecting to see :)
Unfortunately, I don't have a ready-to-use example, it's something to dig into.

I guess `go-releaser` is the most flexible and functional option nowadays, it's supports cosign (since you mentioned it) and so many different options. We can either try to go through their documentation or look at how other projects are using it.

NissesSenap: I can see if I get time to take a look at it today. We could start to merge this PR and then we can work on making it faster, we don't do releases that often so it shouldn't be a big issue. I will also check a bit on our old setup because it don't take 40 min...

NissesSenap: I don't get it, our old repo it only took about 6 min to build the container images: https://github.com/grafana-operator/grafana-operator/actions/runs/4142896346/jobs/7164110490
Sure there are some differences between the setups but it's more or less the same...

helm/grafana-operator/values.yaml
```diff
@@ -0,0 +1,75 @@
...
+# If the operator should run in namespace-scope or not,
+# if true the operator will only be able to manage instances in the same namespace
+namespaceScope: false
+
+# If running multiple replicas of the operator you need to add leaderElect,
+# in general we don't recommend this feature and it's not tested.
+leaderElect: false
+
+image:
+  repository: quay.io/grafana-operator/grafana-operator
```
NissesSenap, line 14: In the example we are currently using ghcr.io how do we want to do it?

Issue #53, opened 2022-12-18
Add status to none grafana CRD if no matching instance is found


NissesSenap:
If I create a dashboard/datasource/folder that don't have any matching instance it's kind of hard to know.

For example with the bellow config i missed the `grafana-a` and I had to look in the logs to find it.
It's kind of painful unless you know how the logs looks like and it would have saved me allot of time to understand what was going on to have it in  the status field.

```.yaml
➜ k get grafanadatasources.grafana.integreatly.org grafanadatasource-sample -o yaml
apiVersion: grafana.integreatly.org/v1beta1
kind: GrafanaDatasource
metadata:
  name: grafanadatasource-sample
  namespace: grafana-operator-system
spec:
  datasource:
    access: proxy
    editable: false
    isDefault: true
    jsonData:
      timeInterval: 5s
      tlsSkipVerify: true
    name: prometheus
    type: prometheus
    url: http://prometheus-service:9090
  instanceSelector:
    matchLabels:
      dashboards: grafana-a
  plugins:
  - name: grafana-clock-panel
    version: 1.3.0
```

NissesSenap:
In this case it was a config error from my side.

```
1.671375627838152e+09	INFO	no matching instances found for dashboard	{"controller": "grafanadashboard", "controllerGroup": "grafana.integreatly.org", "controllerKind": "GrafanaDashboard", "GrafanaDashboard": {"name":"grafanadashboard-sample","namespace":"grafana-operator-system"}, "namespace": "grafana-operator-system", "name": "grafanadashboard-sample", "reconcileID": "3aa203b4-8968-499d-a1c3-365b49c6b87f", "dashboard": "grafanadashboard-sample", "namespace": "grafana-operator-system"}
```

No matching instance found.
It would probably be nice if we add that to a status field to make it easier for our users to find issues like that quickly.

Issue #24, opened 2022-09-13
api/v1beta1/dashboard_list.go strange file name


NissesSenap:
When looking at the functions inside dashboard_list.go i notice that they are general purpose for the whole v1beta1 package.
We should probably change it's name to make it a bit easier to understand that.

And we should write it using more generic variables to show that it's generic used by multiple controllers.

pb82:
resolved in #37

HubertStefanski:
@NissesSenap Seems like you closed and reopened it on the same day? was this intended?

NissesSenap:
Yes, this was in a PR that was abandoned so i don't think it's fixed yet.

PR #75, opened 2023-03-04
Minor fixes before rc0

NissesSenap:
- Fixes: #24
- Remove unused constants
- Remove all usage of `grafana.spec.client.preferIngress:  true`, this is only for development. We don't want our users to use it.
- Update CONTRIBUTING.md on a possible way of working with kind and local development.
- Set `instanceSelector` to required
- Minor yaml fixes
- Fixes: #53


controllers/reconcilers/grafana/deployment_reconciler.go
```diff
@@ -20,10 +20,6 @@ import (
 )
 
 const (
-	InitMemoryRequest                       = "128Mi"
-	InitCpuRequest                          = "250m"
-	InitMemoryLimit                         = "512Mi"
-	InitCpuLimit                            = "1000m"
```
NissesSenap, line 26: These values aren't used so I removed them.

NissesSenap (commented):
My own thoughts about the PR.

controllers/datasource_controller.go
```diff
@@ -164,19 +165,25 @@ func (r *GrafanaDatasourceReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
 
 	if len(instances.Items) == 0 {
 		controllerLog.Info("no matching instances found for datasource", "datasource", datasource.Name, "namespace", datasource.Namespace)
+		datasource.Status.NoMatchingInstances = true
+		err = r.Client.Status().Update(ctx, datasource)
+		if err != nil {
+			controllerLog.Error(err, "unable to update status", "name", datasource.Name)
+		}
+		return ctrl.Result{RequeueAfter: RequeueDelay}, err
 	}
```
weisdd, line 188: Should we add `datasource.Status.NoMatchingInstances = false` somewhere around this line?

controllers/datasource_controller.go
```diff
@@ -327,6 +334,7 @@ func (r *GrafanaDatasourceReconciler) onDatasourceCreated(ctx context.Context, g
 
 func (r *GrafanaDatasourceReconciler) UpdateStatus(ctx context.Context, cr *v1beta1.GrafanaDatasource) error {
 	cr.Status.Hash = cr.Hash()
+	cr.Status.NoMatchingInstances = true
```
weisdd, line 337: Could you, please, explain why this line is added here and not inside a function that checks for presence of matching instances? (assuming `UpdateStatus` can be called from anywhere)
Btw, I see that in different controllers it is set to different values.

controllers/grafanafolder_controller.go
```diff
@@ -169,21 +169,25 @@ func (r *GrafanaFolderReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
 	if len(instances.Items) == 0 {
 		controllerLog.Info("no matching instances found for folder")
-		return ctrl.Result{Requeue: false}, nil
+		folder.Status.NoMatchingInstances = true
+		err = r.Client.Status().Update(ctx, folder)
+		if err != nil {
+			controllerLog.Error(err, "unable to update status", "name", folder.Name)
+		}
+		return ctrl.Result{RequeueAfter: RequeueDelay}, err
 	}
```
weisdd, line 192: Should we add `folder.Status.NoMatchingInstances: false` somewhere around this line?

controllers/grafanafolder_controller.go
```diff
@@ -344,6 +348,7 @@ func (r *GrafanaFolderReconciler) onFolderCreated(ctx context.Context, grafana *
 
 func (r *GrafanaFolderReconciler) UpdateStatus(ctx context.Context, cr *v1beta1.GrafanaFolder) error {
 	cr.Status.Hash = cr.Hash()
+	cr.Status.NoMatchingInstances = false
```
weisdd, line 351: Could you, please, explain why this line is added here and not inside a function that checks for presence of matching instances? (assuming `UpdateStatus` can be called from anywhere)
Btw, I see that in different controllers it is set to different values.

controllers/dashboard_controller.go
```diff
@@ -170,23 +170,26 @@ func (r *GrafanaDashboardReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
-
+		dashboard.Status.NoMatchingInstances = true
+		err = r.Client.Status().Update(ctx, dashboard)
+		if err != nil {
+			controllerLog.Error(err, "unable to update status", "name", dashboard.Name)
+		}
 		// TODO when a label selector has been updated to no longer match any Grafana instances, should we delete the dashboard from those instances?
-		return ctrl.Result{Requeue: false}, nil
+		return ctrl.Result{RequeueAfter: RequeueDelay}, err
 	}
```
weisdd, line 194: Should we add `dashboard.Status.NoMatchingInstances: false` somewhere around this line?

controllers/dashboard_controller.go
```diff
@@ -410,6 +413,7 @@ func (r *GrafanaDashboardReconciler) fetchDashboardJson(dashboard *v1beta1.Grafa
 
 func (r *GrafanaDashboardReconciler) UpdateStatus(ctx context.Context, cr *v1beta1.GrafanaDashboard) error {
 	cr.Status.Hash = cr.Hash()
+	cr.Status.NoMatchingInstances = false
```
weisdd, line 416: Could you, please, explain why this line is added here and not inside a function that checks for presence of matching instances? (assuming `UpdateStatus` can be called from anywhere)
Btw, I see that in different controllers it is set to different values.

controllers/dashboard_controller.go
```diff
@@ -170,23 +170,26 @@ func (r *GrafanaDashboardReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
 	}
 
 	if len(instances.Items) == 0 {
 		controllerLog.Info("no matching instances found for dashboard", "dashboard", dashboard.Name, "namespace", dashboard.Namespace)
-
+		dashboard.Status.NoMatchingInstances = true
+		err = r.Client.Status().Update(ctx, dashboard)
+		if err != nil {
+			controllerLog.Error(err, "unable to update status", "name", dashboard.Name)
+		}
```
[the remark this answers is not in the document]
weisdd: What will happen if there are no matching instances found? Will it be an error or just an empty list? If it's an empty list, then I guess we should then merge two sections (`if err != nil || len(instances.Items) == 0`).

weisdd (changes requested):

CONTRIBUTING.md
```diff
@@ -6,6 +6,77 @@ Thank you for investing your time in contributing to our project.
...
+```shell
+kubectl apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/main/deploy/static/provider/kind/deploy.yaml
+```
+
+Get your kind IP.
+
+```shell
+KIND_IP=$(docker container inspect kind-control-plane \
+  --format '{{ .NetworkSettings.Networks.kind.IPAddress }}')
+sudo sh -c "echo $KIND_IP test.io >> /etc/hosts"
```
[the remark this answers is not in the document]
NissesSenap: I haven't used it but it do sound better. Could you do a commit with some instructions? I'm more then happy to change to a new way in my own flow as well.

CONTRIBUTING.md
```diff
@@ -6,6 +6,77 @@ Thank you for investing your time in contributing to our project.
...
+metadata:
+  name: grafana
+  labels:
+    dashboards: "grafana"
+spec:
+  client:
+    preferIngress: true
+```
+
+This makes the grafana client in the operator to use the ingress address instead of the service name.
```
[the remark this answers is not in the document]
NissesSenap: That is the way we have gotten it to work when running the operator locally. If you have a better solution i totally agree that we should go with a service instead

controllers/dashboard_controller.go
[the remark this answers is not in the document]
NissesSenap: It should have already returned but due to no matches, but I will move the len check up to the other error just to be sure and I have removed the other section

controllers/grafanafolder_controller.go
NissesSenap: I remove it instead.

controllers/dashboard_controller.go
NissesSenap: I have moved in to a separate function witch handles that logic. 
Please have a look now

controllers/reconcilers/grafana/deployment_reconciler.go
HubertStefanski: :+1: they must have been ported over from v4, no need to keep them here

HubertStefanski:
I'm happy with this MR if @weisdd approves based on your conversation. 

Nice simplification of the logic :+1:

controllers/datasource_controller.go
```diff
@@ -401,3 +392,22 @@ func (r *GrafanaDatasourceReconciler) SetupWithManager(mgr ctrl.Manager, ctx con
 
 	return err
 }
+
+func (r *GrafanaDatasourceReconciler) GetMatchingDatasourceInstances(ctx context.Context, datasource *v1beta1.GrafanaDatasource, k8sClient client.Client) (v1beta1.GrafanaList, error) {
+	instances, err := GetMatchingInstances(ctx, k8sClient, datasource.Spec.InstanceSelector)
+	if err != nil || len(instances.Items) == 0 {
+		datasource.Status.NoMatchingInstances = true
+		err = r.Client.Status().Update(ctx, datasource)
```
weisdd, line 400: I think this will be a bit inconsistent in its behavior:
`err` generated by `GetMatchingInstances` is never returned from the function as it gets overridden by the status update. So, successful status update erases the error while failing status update makes the caller think as if there are no matching instances.
The question is whether in this helper you want to return status update errors at all or better to just log it while returning only errors directly related to the matching instances lookup.

I haven't thought it out fully, but if we always consider having no matching instances as an error, maybe it'll make sense to embed this logic into `GetMatchingInstances` (e.g. define a custom error for that and return it).

CONTRIBUTING.md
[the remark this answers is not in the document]
weisdd: Ah, I didn't know you have a different flow. - I just deploy the operator to the cluster and never run it from local machine. :)
So, alright, assuming you run the operator locally, how does it know how to connect to k8s?
Also, I don't see test.io in grafana's spec. How does it know the ingress name then?

weisdd (changes requested):

CONTRIBUTING.md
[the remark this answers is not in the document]
NissesSenap: I have updated the example with the test.io so it makes more seance. 
I think this should be a decent start but I would love if you can add your way of working as an example as well in a separate PR.

controllers/datasource_controller.go
NissesSenap: Yeah I was thinking about this as well.
I guess one of the questions we have to ask our self is why wouldn't the operator be able to update the status? Is it because it was removed or access rights? Is there any other reasonable reason?

But as you say it's probably better to give the user the correct information that there actually is an hit.
We probably should go over to custom errors but it's probably something that we should do everywhere.
So lets do it in a separate PR and see if we can get this merged if it looks okay with you?

weisdd: The code still overwrites the error. For status updates errors, we need to either use another variable name or use the syntax that scopes err to the if structure, e.g.:

```go
func (r *GrafanaDashboardReconciler) GetMatchingDashboardInstances(ctx context.Context, dashboard *v1beta1.GrafanaDashboard, k8sClient client.Client) (v1beta1.GrafanaList, error) {
	instances, err := GetMatchingInstances(ctx, k8sClient, dashboard.Spec.InstanceSelector)
	if err != nil || len(instances.Items) == 0 {
		dashboard.Status.NoMatchingInstances = true
		if err := r.Client.Status().Update(ctx, dashboard); err != nil {
			r.Log.Info("unable to update the status of %v, in %v", dashboard.Name, dashboard.Namespace)
		}
		return v1beta1.GrafanaList{}, err
	}
	dashboard.Status.NoMatchingInstances = false
	if err := r.Client.Status().Update(ctx, dashboard); err != nil {
		r.Log.Info("unable to update the status of %v, in %v", dashboard.Name, dashboard.Namespace)
	}

	return instances, err
}
```

NissesSenap: Thanks, my brain wasn't really working...
I use the code you suggested, thanks @weisdd

weisdd (approved):

pb82 (approved):

PR #79, opened 2023-03-09
docsy hugo site

NissesSenap:
- Add contribution instructions for hugo.
- Adapt docs to work with hugo
- Setup docsy hugo page
- Create CI/CD solution to deploy homepage


.github/workflows/e2e.yaml
```diff
@@ -11,8 +11,34 @@ env:
   NAMESPACE: grafana-operator-system
 
 jobs:
+  docs_only_check:
```
NissesSenap, line 14: Create a check to not run e2e test if there is only documentation changes.

docs/about/_index.html
```diff
@@ -0,0 +1,39 @@
...
+  main:
+    weight: 10
+---
+
+{{< blocks/section >}}
+
+<div class="col-12">
+<h1 class="text-center">About Grafana-Operator</h1>
+
+<h2>Creation</h2>
```
NissesSenap, line 14: This is the feeling I have gotten but I don't know the official version. But I guess I wanted to show a bit where the operator came from and where we are going.

docs/docs/proposals/001 external grafana instance.md
```diff
@@ -1,18 +1,22 @@
...
 This includes design elements to support the following with Grafana operator :
+
 - Ability to define an external Grafana instance as Grafana source to Grafana operator
 - Ability to create Grafana dashboards on a remote Grafana Instance.
 - Ability to add external cloud data sources to an external Grafana instance.
 
 ## Info
 
-status: Draft
+status: Implemented
```
NissesSenap, line 19: This have actually been implemented so why not fix it.

NissesSenap (commented):
I think html and design is about as fun as I show you with this page.
I have copied more or less everything from https://github.com/google/docsy-example and made a few minimal changes.

From my point of view this is a way of getting started with hugo and hopefully being able to have a place for people to look at docs that isn't in a github README.

I would love that someone that is better and think it's more fun then I do to actually make something that looks good.

Added a few comments for clarity.
Looking forward to your thoughts.

.github/workflows/e2e.yaml
NissesSenap: This won't work during this PR due to a few files is .yaml. And I do want to e2e test yaml changes.

HubertStefanski (approved):

PR #80, opened 2023-03-10
remove spec.grafanaContainer and add example to define custom image

### LINKED ISSUE #77: `spec.grafanaContainer.baseImage` unused
We've noticed, that the field `spec.grafanaContainer.baseImage` goes unused during reconciliation of the Deployment. 

Although, it is possible to override the image via
```yaml
spec:
  deployment:
    spec:
      template:
        spec:
          containers:
            - name: grafana
              image: harbor.registry/grafana:1234
```
But that's kinda a lot of overhead, just to overwrite the image. So I hope we can keep the baseImage parameter (as it's already in the current v4 as well) :)

There's also the `initImage` parameter right next to it. But I guess that's obsolete now?

tamcore:
This will remove obsolete `grafanaContainer` references and add an example to showcase referencing a custom image. Fixes #77.


HubertStefanski (approved):
Thanks @tamcore ! LGTM

NissesSenap:
LGTM, thanks allot @tamcore

PR #82, opened 2023-03-10
Fix Custom Folder Logic in Dashboard Controller

### LINKED ISSUE #84: Deploying a dashboard with a custom folder doesn't work
Deploying a folder with the field:
```yaml
spec:
   folder: my-folder
```

Creates the folder on the grafana API of the matching grafana instance, however, the dashboard is never assigned to the folder, resulting in an empty 

Example of bug:

```yaml
---
kind: Grafana
metadata:
  name: grafana
  labels:
    dashboards: "grafana"
spec:
  client:
    preferIngress: true
  config:
    log:
      mode: "console"
    auth:
      disable_login_form: "false"
    security:
      admin_user: root
      admin_password: secret
---
kind: GrafanaDashboard
metadata:
  name: grafanadashboard-from-url-with-folder
spec:
  folder: "my-custom-folder"
  instanceSelector:
    matchLabels:
      dashboards: "grafana"
  url: "https://raw.githubusercontent.com/integr8ly/grafana-operator/master/deploy/examples/remote/grafana-dashboard.json"
```

HubertStefanski:
Fixes:
- Wrong field assignment in grafana API dashboard struct, resulting in folders not being assigned to the correct folder
- Change `deleteFolderIfEmpty` method to use UID instead of ID, according to https://grafana.com/docs/grafana/v8.4/http_api/folder/ The delete function uses a UID, rather than ID, as previously used.

Adds:
- Example dashboard_with_custom_folder deployment

Fixes #84

### DIFF
--- a/controllers/dashboard_controller.go
+++ b/controllers/dashboard_controller.go
@@ -22,7 +22,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
-	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


pb82 (approved):
Verified locally, example dashboard ends up in the custom folder :+1:

PR #87, opened 2023-03-11
fix: remove command override from manifests

weisdd:
- Removed the command (=entrypoint) override from manifests:
  - When an image is built with ko, the binary is named differently, so `make deploy` will fail due to the command override in the manifests. I think we should not have it in manifests at all, the choice of binary name is up to a particular image whether it's build through existing Dockerfile or with `ko`.


NissesSenap (approved):
Agreed, I rebased and added the same fix to the helm chart.

PR #88, opened 2023-03-11
feat: use ko in kuttl e2e

weisdd:
- Moves from docker builds to ko for e2e tests.

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -3,8 +3,6 @@ name: KinD End-to-End
 on:
   pull_request:
     branches: [master]
-  push:
-    branches: [master]
 
 env:
   NAME: grafana-operator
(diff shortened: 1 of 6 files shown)


.github/workflows/e2e.yaml
```diff
@@ -18,36 +16,39 @@ jobs:
...
-      - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v2.4.1
+          version: v0.13.0
+
       - name: Set up KinD
         id: kind
         run: |
           kind create cluster --image=kindest/node:v1.25.3 --config tests/e2e/kind.yaml
+
+      - name: Set up Docker Buildx
```
NissesSenap, line 34: Why is docker buildx still needed?

NissesSenap:
Should we remove the container image as well? Or we can do that in a separate PR where we also update the contribution guide.

.github/workflows/e2e.yaml
weisdd: When I tried to make `ko` load image directly to kind on mac, it failed, so just made it load image to docker daemon and then explicitly pass it to kind `kind load docker-image`. Direct load might work on runner / Linux, IDK.
We don't need buildx per say, pure docker is enough as well.

weisdd:
@NissesSenap Yeah, I think that it's better to leave `Dockerfile` until we have an up-to-date contribution guide.

.github/workflows/e2e.yaml
NissesSenap: Okay, well It looks like it works. Really nice, I will merge it

PR #89, opened 2023-03-12
Developer experience and helm/kustomize consistency

NissesSenap:


pb82:
Thanks @NissesSenap this integrates `gofumpt` nicely. When running `make code/golangci-lint` it looks like the linter has problems with inherited fields:

```
api/v1beta1/grafanadashboard_types.go:228:16: dashboard.Namespace undefined (type GrafanaDashboard has no field or method Namespace) (typecheck)
                if dashboard.Namespace == namespace && dashboard.Name == name {
```

This complains about the field `Namespace` which is not directly a member of the `GrafanaDashboard` struct, but rather a member of `ObjectMeta` from which `GrafanaDashboard` inherits.

How can we convince the linter that this is ok? There are many other errors of this kind? Or is it an issue with the version of golangci-lint that i'm running (1.41.1)?

pb82 (approved):

NissesSenap:
Spoken to @pb82 and the issue was an older golangci-lint already installed on he's computer. It now works as expected for him.

PR #90, opened 2023-03-12
Change from docker to KO everywhere

NissesSenap:
We are no longer using docker as our image solution and this PR will match that.

- make add ko
- make remove docker
- Write basi contributor docs with ko
- Remvoe Dockerfile


Makefile
```diff
@@ -193,73 +165,13 @@ BUNDLE_DEFAULT_CHANNEL := --default-channel=$(DEFAULT_CHANNEL)
...
-.PHONY: bundle-build
-bundle-build: ## Build the bundle image.
-	docker build -f bundle.Dockerfile -t $(BUNDLE_IMG) .
-
-.PHONY: bundle-push
-bundle-push: ## Push the bundle image.
-	$(MAKE) docker-push IMG=$(BUNDLE_IMG)
-
-.PHONY: opm
-OPM = ./bin/opm
```
NissesSenap, line 223: Don't know what a bunch of this is, to my knowledge we only need the `make bundle` command to generate all the files needed by OLM. Is any of these commands something that you use in your daily lifes @pb82 @HubertStefanski ?

NissesSenap (commented):
This isn't finished but it's a cleanup at least.

CONTRIBUTING.md
```diff
@@ -27,6 +27,12 @@ make code/golangci-lint
 
 Depending on what you have changed these commands will update a number of different files.
 
+### KO
+
+To speed up the container build process and not being forced to maintain a Dockerfile we are using [ko](https://ko.build/).
```
weisdd, line 32:
```suggestion
To speed up multi-arch image builds and avoid burden of maintaining a Dockerfile, we are using [ko](https://ko.build/).
```
Just to highlight what exactly we want to speed up :)

Makefile
```diff
@@ -308,6 +220,27 @@ endif
...
+# If you want to push ko to your local Docker daemon
+.PHONY: ko-build-local
+ko-build-local: ko
+	KO_DOCKER_REPO=ko.local/grafana-operator/grafana-operator $(KO) build  --sbom=none --bare
+
+# If you want to push ko to your kind cluster
+.PHONY: ko-build-kind
+ko-build-kind: ko
+	KO_DOCKER_REPO=kind.local $(KO) build --sbom=none
+
```
weisdd, line 243:
```suggestion
export KO_DOCKER_REPO := ko.local/grafana-operator/grafana-operator

# If you want to push ko to your local Docker daemon
.PHONY: ko-build-local
ko-build-local: ko
	$(KO) build --sbom=none --bare

# If you want to push ko to your kind cluster
.PHONY: ko-build-kind
ko-build-kind: ko
	$(KO) build --sbom=none --bare
	kind load docker-image $(KO_DOCKER_REPO)
```

weisdd (changes requested):

Makefile
pb82: opm is a CLI to build catalog images. I wouldn't say I use it daily, but we will need it for the OLM releases.

NissesSenap: But I just use `make bundle` to generate the code that I then push up to the central repositories, but I might have been doing it wrong all this time ^^.

But if you use it I will of course not remove it.

pb82: You don't need it for the bundle, only for the catalog image. Which you don't usually build yourself if you release to operatorhub.io, etc. It is still useful though for testing to be able to build our own catalog images.

PR #91, opened 2023-03-12
Change from quay.io to ghcr.io

NissesSenap:
ghcr is the current repo location that we use,
in the future we might support multiple repositorieis.


NissesSenap:
My yaml formater gave some files some love but in general the only changes made is quay.io -> ghcr.io

weisdd (approved):
LGTM

PR #92, opened 2023-03-13
Individual helm chart versioning

nissessenap:
This to be able to support individual helm chart versioning.

### DIFF
--- a/.github/workflows/release-github.yaml
+++ b/.github/workflows/release-github.yaml
@@ -24,9 +24,10 @@ jobs:
         run: |
           cd deploy/helm
           yq -i '.image.repository = "${{ env.GHCR_REPO }}"' grafana-operator/values.yaml
+          HELM_CHART_VERSION=$(cat grafana-operator/Chart.yaml | yq -r '.version')
           helm registry login -u ${{ github.actor }} -p ${{ secrets.GITHUB_TOKEN }} ghcr.io
-          helm package --app-version ${{ github.ref_name }} --version ${{ github.ref_name }} grafana-operator
-          helm push grafana-operator-${{ github.ref_name }}.tgz ${{ env.HELM_REPO }}
+          helm package --app-version ${{ github.ref_name }} grafana-operator
+          helm push grafana-operator-$HELM_CHART_VERSION.tgz ${{ env.HELM_REPO }}
 
   image:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


deploy/helm/grafana-operator/Chart.yaml
```diff
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.1.1
+version: 0.1.0-rc0
```
NissesSenap, line 18: This would force us to make changes like this before cutting a new release. But we will still need to make prepare PR:s just like we do in our current workflow so I don't think it's a huge problem.

.github/workflows/release-github.yaml
```diff
@@ -70,3 +71,4 @@ jobs:
             --image-label org.opencontainers.image.revision=${{ github.sha }} \
             --image-label org.opencontainers.image.version=${{ github.ref_name }} \
             --image-label org.opencontainers.image.created=${{ env.BUILD_DATE }} \
+            --insecure-registry
```
NissesSenap, line 74: I assumed this would get fixed quicker, I was wrong...

.github/workflows/release-github.yaml
```diff
@@ -24,9 +24,10 @@ jobs:
         run: |
           cd deploy/helm
           yq -i '.image.repository = "${{ env.GHCR_REPO }}"' grafana-operator/values.yaml
+          HELM_CHART_VERSION=$(cat grafana-operator/Chart.yaml | yq -r '.version')
```
NissesSenap, line 27: This is the actually interesting change of this PR

nissessenap (commented):
So I feel it would be nice to have individual versioning of the helm chart.
But with the current setup we will never be able to trigger a new release of the helm chart unless we release a new tag and that will release a new container image.

So the question is should we have individual chart versioning?
If so is it worth doing it if we are unable to release anything except when we do a normal release of the operator?
If we don't think so should we instead create some kind of tag schema.
If you tag with helm/* then trigger the helm chart pipeline and not else.
This will force us to trigger two releases.

Is that a problem?

NissesSenap:
Lets talk about this after we merge rc0 to the main repo.

.github/workflows/release-github.yaml
weisdd, line 27:
```suggestion
          HELM_CHART_VERSION=$(yq -r '.version' grafana-operator/Chart.yaml)
```
That's a bit simpler :)

